PR #1, opened 2018-06-19
Mixed-Content fix for MDI icons when using TileBoard over HTTPS

cgarwood:
Fixes a mixed-content warning when loading the MDI icons if you're running TileBoard over HTTPS


resoai:
We had :// at the very beginning but this would not work with file:// proto. I have changed the URL to https which should fix it.

PR #8, opened 2018-06-20
Add suport for @/& prefixes in subtitles

vpnmaster:


Evgeny-:
Thanks for the commits! We've just released new iteration with prefixes support for titles and subtitles.

PR #45, opened 2018-06-27
Fix sensor tile to show default unit_of_measurement

cgarwood:
The sensor tile would only show a unit if `unit:` was defined in the JS configuration. This fix allows it to show the default unit_of_measurement from HomeAssistant if a `unit:` is not defined in JS.


cgarwood:
Gah, edited an old version of the file. Lots of progress on this. I'll rebase and open a new PR

PR #46, opened 2018-06-27
Fix for sensor tile unit_of_measurement

cgarwood:
The sensor tile would only show a unit if `unit:` was defined in the JS configuration. This fix allows it to show the default unit_of_measurement from HomeAssistant if a `unit:` is not defined in JS.

### DIFF
--- a/tile.html
+++ b/tile.html
@@ -51,7 +51,7 @@
       <div class="item-entity">
          <span ng-bind="entityValue(item, entity)"
                class="item-entity--value"></span>
-         <span ng-if="item.unit" class="item-entity--unit"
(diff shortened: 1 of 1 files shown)


resoai:
I'm not sure about this one as this would disallow leaving unit blank.

Evgeny-:
What about `ng-if="item.unit !== false"` ?

Evgeny-:
Or make a function for this

cgarwood:
Not super familiar with Angular, but couldn't you just set `unit: ''` if you wanted to hide the unit? Or does Angular equate that to null/false?

Evgeny-:
Not an angular but javascript itself.

PR #54, opened 2018-06-30
Add support for custom stylesheet

cgarwood:
Includes a styles/custom-styles.css in index.html (and adds it to .gitignore) so custom CSS styles can be added and not lost when updating from the repository.


Evgeny-:
ok, why not

Evgeny-:
only filename is debatable, maybe just `custom.css` ?

PR #77, opened 2018-07-08
Add favicon

gerard33:
The html page was missing a favicon, so adding one from https://materialdesignicons.com.


gerard33:
I just took this dashboard icon https://materialdesignicons.com/icon/view-dashboard.
But of course another icon is fine as well, just missing one in the current page.

resoai:
If I remember correctly, in order to make it correctly, there must be about 10 different sizes for iOS and about 5 for Android :) I'd love to have something unique for the icon, so let's get back to this one.

gerard33:
Maybe this link is helpful https://gist.github.com/ianaya89/45731a1ab52a9fb6e7d9.
Now to find a nice icon :grinning:

resoai:
Thanks, I have that list. The problem is to come up with a decent icon which is proving to be difficult.

Evgeny-:
Ok, i've added some favicon. Probably it isn't perfect, but unless some one shows something better it will be like that.

Issue #78, opened 2018-07-08
Feature request: Custom image for weather icon


rchl:
As far as I can tell, we can only use mdi icons for weather icon. The API I'm using provides its own image and there can be a lot of variants of the icon (see https://om.yr.no/symbol/) so it's hard to map all of those to mdi icons. So I would suggest having support for specifying own icon image instead of using mdi icons.

rchl:
Just realized that not mdi icons are used but included svg icons. Still, it would be a lot of work to try to map those to ones provided by `yr` and probably some wouldn't match exactly.

resoai:
We are actually using Weather Underground icons with images stored in `\images\weather-icons`.

resoai:
So what does YR sensor return? What is the value of the icon?

rchl:
It returns numeric value of the icon in `state` (for example `sensor.yr_symbol.state: 2`) and corresponding URL of the icon image in attributes:
```
attribution: Weather forecast from met.no, delivered by the Norwegian Meteorological Institute. friendly_name: YR Symbol
entity_picture: https://api.met.no/weatherapi/weathericon/1.1/?symbol=2;content_type=image/png
```
I could take a shot at implementing this if you agree it is useful and we can agree on how it should be defined in the config file.

rchl:
I'm not claiming that YR's images are better visually but if the choice is to either do the manual labor of mapping YR's icons to bundled ones (and as I said, there would probably be no 1=1 matches for some icons -- but possibly I'm wrong here) or just using whatever API provides, then I would think that latter is easier and more future proof.

YR does actually also provides icons in SVG format so it would work in all sizes.
For example: https://www.yr.no/grafikk/sym/v2016/svg/01d.svg

There is more information about the API here: https://om.yr.no/info/verdata/spesifikasjon/ (in Norwegian).

Unfortunately the HA component (https://www.home-assistant.io/components/sensor.yr/) doesn't expose URLs of those "new" images (there is "old" and "new" variant of icons), so it would be hard to use those SVG images currently. The component would need to be updated first...

resoai:
We could create a CSS file with classes for YR icons. Let's think about it.

rchl:
> We could create a CSS file with classes for YR icons. Let's think about it.

Sure but that would be catering for one specific sensor instead of having generic solution.

> Couldn't you just use the .attributes.entity_picture of the YR sensor as the icon on the weather tile?

[...]

PR #80, opened 2018-07-10
Feature/weather icon

rchl:
Implement support for specifying image URL for weather icon (#78)


rchl:
Note that I've used template literals in one place so this can break older browsers (IE11). Not sure if the aim is to support that one.

Evgeny-:
Doesn't seem like "url:" prefix is a good solution

rchl:
Other options I can think of is to have either completely separate set of properties (iconUrl, iconUrlMap) or one extra property (e.g. iconAsUrl: true/false) which would change behaviour of icon/iconMap.

(Also, URL relative to server URL should probably be handled too)

Evgeny-:
I think `icon` and `iconUrl` would be ok. iconMap we deprecate and move them to `icons`. `iconUrl` would have greater priority.

scripts/directives.js
```diff
@@ -257,3 +257,46 @@ App.directive('iframeTile', ['$interval', function ($interval) {
...
+      `,
+      link: function ($scope, $el, attrs) {
+         $scope.iconClass = '';
+         $scope.iconStyle = {};
+
+         var updateIcon = function () {
+            var iconValue = $scope.getWeatherIcon({
+               item: $scope.item,
+               entity: $scope.entity,
+            });
```
Evgeny-: Is it a mistake? Calling function with object as argument?

scripts/directives.js
```diff
@@ -257,3 +257,46 @@ App.directive('iframeTile', ['$interval', function ($interval) {
       }
    }
 }]);
+
+App.directive('weatherIcon', [function () {
```
Evgeny-: Not sure if we really need directive

scripts/directives.js
rchl: In angularjs, when function is bound to a scope, it has to be called that way. The object must have keys that match the signature at the bound-site.

Evgeny-: Oh, it's isolated (&). Well anyway, i'm think we don't really need directive here.

scripts/directives.js
Evgeny-: We can just separate elements, i'll try to do it.

Evgeny-:
Somehow implemented

PR #83, opened 2018-07-11
Add custom HTML to screensaver & hideOverlay capability

cgarwood:
Allows specifying custom HTML to be included on a screensaver slide, as well as hiding the clock overlay.

hideOverlay default to false if not specified.

```
slides: [
	{
		bg: 'images/bg3.jpg',
		html: '<div class="test">This is a test</div>',
		hideOverlay: true
	},
	{
		bg: 'images/slide2.jpg'
	}
]
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -250,10 +250,12 @@
    <div class="screensaver-slides" ng-if="isShowed">
       <div class="screensaver-slide" ng-style="getSlideStyle(slide)"
            ng-class="getSlideClasses($index, slide)"
-           ng-repeat="slide in slides track by $index"></div>
+           ng-repeat="slide in slides track by $index">
+           <div class="screensaver-html" ng-if="slide.html" ng-bind-html="slide.html"></div>
+      </div>
    </div>
 
-   <div class="screensaver-content" ng-if="isShowed">
+   <div class="screensaver-content" ng-if="isShowed && !hideOverlay">
       <div class="screensaver-datetime">
          <div class="screensaver-time">
             <clock></clock>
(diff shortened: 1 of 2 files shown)


Evgeny-:
Do we really need this?

resoai:
Hiding clock is certainly not something which should be allowed on a slide level. What is the use case for text? :)

cgarwood:
In my previous screensaver setup with HADashboard I had a weather slide (current conditions & forecast), and if weather alerts were issued, an alert slide with the text of the alert. Also planned an "Upcoming Events" or calendar slide but never got around to implementing it. Since those were more text/informational, hiding the default clock (and weather overlay in my HADashboard setup) was beneficial.

resoai:
I would think that if something besides clock is to be shown, it should not be bound to a particular slide but be independent i.e. global.

cgarwood:
I think it makes sense for slides to be able to hide the overlay, if you're getting into slides that are more than just photos (like a weather/weather alert slide, or calendar slide). Outside of that, I don't think it's necessary for slides to change what content is on the overlay, other than a blanket "hide the overlay" or "show the overlay (default)"

Most people probably won't use the advanced features, but it doesn't really add any overhead and adds a lot of customization abilities for those who would want to get more advanced.

Evgeny-:
This is going too far. It's ok if we will have custom html not for slide but for whole screensaver, but hiding clocks inside slides is not really a health thing

Evgeny-:
Implemented. Thanks

PR #85, opened 2018-07-12
Fix subtitle text overflowing

rchl:
Saw that with long media title in media player tile


rchl:
There are some small changes in css that are due to, I think, using different (newer) version of less compiler. I've used version:
```
% lessc --version
lessc 3.7.1 (Less Compiler) [JavaScript]
```
which I've installed with `npm install -g less`.

It seems doing a bit better job than your version. Also not including deprecation notice in css. Maybe try upgrading your version?

PR #91, opened 2018-07-15
Added optional date to header

NilSkilz:
I've added the date to my header, which is controlled by an optional parameter in the config JSON. 

I felt like you may have had plans for this yourself? based on the clockStyles() function? Anyway it could probably be expanded to allow for right aligned clocks/dates etc. It would be cool to know if you had any plans yourself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,15 @@ var CONFIG = {
       /* timeFormat: 12 for AM/PM marker, 24 for 24 hour time (default) */
       timeFormat: Number,
 
+      /* showDate: true shows the date in the header, false hides it (default) */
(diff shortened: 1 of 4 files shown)


Evgeny-:
Just do not include header in the page

NilSkilz:
I'm confused by this comment. How does that achieve the objective of having the date and time on the screen?

Evgeny-:
Oh. i see you did opposite. My mistake, sorry. 

We should work on header, may be will be better open an issue.

Evgeny-:
Here it is https://github.com/resoai/TileBoard/issues/92

PR #93, opened 2018-07-16
Add iOS full-screen mode

gerard33:
On iOS the tileboard can now be added to the frontscreen at full-screen mode.
Code is from https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html.


resoai:
Thank you, I've merged it!

Issue #100, opened 2018-07-19
menu on the bottom in mobile layout overlaps tiles


rchl:
When:
 - Using mobile layout (`customTheme: CUSTOM_THEMES.MOBILE`)
 - Having menu on the bottom (`menuPosition: MENU_POSITIONS.BOTTOM`)
 - There is enough tiles to enable page scrolling
 - Page is scrolled to the very bottom

then the bottom menu partially overlaps bottom tiles which, depending on the overlapped tiles, might make it harder/impossible to interact with them.

See a screenshot:
![mobile-toolbar](https://user-images.githubusercontent.com/153197/42924778-013afaaa-8b2c-11e8-87ad-c85e3a4b2d22.png)

It should be possible to scroll more so that tile is not overlapped by menu.

Evgeny-:
Just add little bit to the height of group. e.g. + 0.5

Evgeny-:
Probably bottom margin of group also should work

resoai:
Presume the matter is resolved and we can close this one?

rchl:
While I appreciate how customisable this project is and how easily we can workaround issues, I think this should still be fixed in core.

It would be easy to fix it for mobile layout, but I think it should be fixed for all themes which would be quite a bit more work. Basically make sure that it's always possible to scroll page so that tiles are unobstructed by toolbar. For non-mobile layouts toolbar is even bigger so it's even easier to make it obstruct tiles if there is enough content vertically.

resoai:
For non-mobile themes which  have page menu on the left, having menu  overlap the tiles is kinda a feature.

rchl:
I'm only talking about menu on the bottom. I haven't seen menu on the left obstructing tiles.

rchl:
OK, I see it. But in that case there is no problem as it's possible to scroll page all the way to the left (or right, depending how you look at it) and then no tiles are obstructed.

resoai:
Would you be able to provide a PR of what you think would be the best fix? Please keep in mind that we should only edit .less.

rchl:
For my specific issue the fix could be:

```diff
diff --git a/styles/themes.less b/styles/themes.less
index 37c1a58..56c282f 100644
--- a/styles/themes.less
+++ b/styles/themes.less
@@ -151,6 +151,18 @@
          }
       }
    }
+
+   &.-menu-bottom {
+      .page {
+         padding-bottom: 44px;
+      }
+
+      @media (min-width: 500px) {
[...]

PR #102, opened 2018-07-21
Fix/page menu obstruction

rchl:
Fix partially obstructed tiles by bottom-positioned page menu (#100)

    When there is enough content to enable vertical scrolling and page menu toolbar
    is at the bottom, the tiles that were at the bottom might have been obstructed
    by the toolbar, making it impossible to click some parts of them.

    Fixed by adding bottom padding to page area equal to size of the toolbar
    (+ toolbar's bottom offset when not 0). This ensures that tiles can be
    scrolled fully in view without being obstructed.

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -255,6 +255,9 @@ body {
   margin-top: 0.1em;
   font-size: 1em;
 }
+.-menu-bottom .page {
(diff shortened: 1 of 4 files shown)


resoai:
Thanks!

Issue #104, opened 2018-07-21
Transparent portions of page menu toolbar should allow click-through


rchl:
If there is enough tiles to trigger either horizontal or vertical scrolling then the tiles might end up under the transparent part of the page menu toolbar and while it looks like it should still be possible to click those tiles, they are obstructed by invisible toolbar and can't be clicked.

In themes that have that toolbar transparent, we should probably use pointer-events none on those regions without buttons.

Here you can see areas highlighted in red where clicking won't trigger action on tiles beneath:
![unclickable-area](https://user-images.githubusercontent.com/153197/43035264-f1aba45a-8cec-11e8-849b-48edc3cde201.png)

rchl:
The fix should not apply themes that don't have transparent toolbar like homekit or winphone.

Evgeny-:
They are shouldn't

rchl:
I don't understand your last comment and why it was closed.

Do you mean that tiles shouldn't be clickable even though they are clearly unobstructed. Don't you think that's quite counter-intuitive?

Evgeny-:
No. Because menu don't suppose to be above tiles. except mobile themes with vertical align.

rchl:
@resoai said that horizontal scrolling is by design and that problem affect horizontal scrolling also. I don't understand why vertical scrolling would be unexpected. It doesn't cost anything to support that
(and if you insist on not supporting it, maybe disable vertical scrolling explicitly).

Evgeny-:
Vertical scroll supported with vertical menu or with mobile theme. Though you still can make pull request with solving this "problem"

rchl:
I've created solution that instead of making that transparent menu area clickable, just darkens the area so that it's clear that the tiles underneath can't be clicked.

It looks like this:
<img width="286" alt="darkened" src="https://user-images.githubusercontent.com/153197/43267315-e95c2a86-90ed-11e8-96dc-1de7d578a4f9.png">

PR #107, opened 2018-07-25
Use padding in bottom-positioned pages-menu instead of bottom offset

rchl:
The only reason for this change is to prepare code for future fix.
It should be almost equally visually to previous code.

Previous code applied bottom offset to bottom-positioned page menu which
created a gap that would let clicks go through it if page was scrolled.
Using padding instead makes sure there is no gap there and also allows
to created better scrolling indicator in the future.

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -256,7 +256,7 @@ body {
   font-size: 1em;
 }
 .-menu-bottom .page {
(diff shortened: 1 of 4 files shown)


rchl:
I know that it seems like an unnecessary change but I have some idea for a fix for #104 and it would help there.

PR #108, opened 2018-07-25
Fix header weather widget image style

rchl:
Image was repeating and wasn't centered as there are no elements with
class .header-weather--icon--image, only .header-weather--icon-image
(--image vs. -image). Move css rules to the correct selector.

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -232,6 +232,9 @@ body {
 }
 .header-weather--icon > div,
 .header-weather--icon-image > div {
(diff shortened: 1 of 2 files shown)


Evgeny-:
It will be better just remove one '-'

rchl:
Sorry, I'm not too familiar with that class naming convention so might have done a poor judgement here.

Evgeny-:
Me neither :) 
It's just chaos

PR #110, opened 2018-07-26
Darken transparent page menu background when page is scrolled (#104)

rchl:
To clearly indicate that the area under page menu is not clickable when
there are tiles underneath, darken its barkgroung when page is scrolled either
vertically (with page menu on the bottom) or horizontally (page menu on the left).

### DIFF
--- a/scripts/directives.js
+++ b/scripts/directives.js
@@ -303,18 +303,32 @@ App.directive('onScroll', [function () {
          onScrollModel: '=',
       },
       link: function ($scope, $el, attrs) {
+         var lastScrolledHorizontally = false;
+         var lastScrolledVertically = false;
+
          var determineScroll = function () {
-            $scope.onScrollModel.scrolledHorizontally =
-               $el[0].scrollLeft !== 0;
-            $scope.onScrollModel.scrolledVertically =
-               $el[0].scrollTop !== 0;
+            var scrolledHorizontally = $el[0].scrollLeft !== 0;
+            var scrolledVertically = $el[0].scrollTop !== 0;
+
+            if(lastScrolledVertically !== scrolledVertically ||
+               lastScrolledHorizontally !== scrolledHorizontally) {
+               $scope.onScrollModel.scrolledHorizontally =
+                  lastScrolledHorizontally = scrolledHorizontally;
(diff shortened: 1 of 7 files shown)


scripts/directives.js
```diff
@@ -295,3 +295,28 @@ App.directive('date', ['$interval', function ($interval) {
...
+               $el[0].scrollLeft !== 0;
+            $scope.onScrollModel.scrolledVertically =
+               $el[0].scrollTop !== 0;
+         };
+
+         determineScroll();
+
+         $el.on('scroll', function () {
+            determineScroll();
+            $scope.$apply();
```
rchl: Not very efficient to trigger digest on each scroll event. I could add throttling here if deemed important.

cgarwood:
Haven't looked at the code, but based on the screenshot I like it. It adds a bit of depth to the page so it's clear that the tiles are behind a menubar.

resoai:
I think it is OK however `on('scroll'` will absolutely destroy it.

Evgeny-:
If you gonna merge it, it will be MUCH better to store previous states of scrolls, and fire digest via `updateView()` only if they are changed.

styles/main.css
```diff
@@ -294,18 +300,33 @@ body {
   bottom: 0;
   top: 0;
 }
+.pages-menu.-left .pages-menu--scroll-indicator {
+  background-image: linear-gradient(to right, rgba(0, 0, 0, 0.1) 98%, rgba(0, 0, 0, 0.3) 98%, rgba(0, 0, 0, 0));
```
Evgeny-: Can we replace all gradients with just more darker border?

rchl: I think darker border wouldn't give this nice 3d effect that helps seeing what's going on. But background color + border shadow should work if you prefer that.

rchl:
BTW. Maybe we should consider always having this darkened background instead of just activating it on scroll? Would simplify the code and looks nice I think.
It's your call though.

rchl:
But then we would need to take header into consideration also. It wouldn't look good if header would overlap the darkened area.

rchl:
I believe so yes.

PR #112, opened 2018-07-26
Added ability to request a specfic page via URL

dlashua:
A URL like:

http://127.0.0.1:8123/local/TileBoard/index.html?blargh

will cause TileBoard to open on a page with an "id" of "blargh".

This is useful if you have mounted panels in multiple rooms and would like each room to start TileBoard on a page devoted to that room.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1552,9 +1552,19 @@ function MainController ($scope) {
          var desiredPage;
 
          if(window.location.search != '') {
-           var desiredPageName = window.location.search.substring(1);
-           var desiredPages = CONFIG.pages.filter(function(obj) {
-             return obj.id == desiredPageName;
+
+           var queryStringPairs = location.search.slice(1).split('&');
+
+           var queryStringParameters = {};
+           queryStringPairs.forEach(function(pair) {
+               pair = pair.split('=');
+               queryStringParameters[pair[0]] = decodeURIComponent(pair[1] || '');
+           });
+
+
+           var desiredPageID = queryStringParameters.pageid
+           var desiredPages = $scope.pages.filter(function(obj) {
+             return obj.id == desiredPageID;
            });
            if(desiredPages.length > 0) {
              desiredPage = desiredPages[0];
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -3,8 +3,6 @@ App.controller('Main', ['$scope', MainController]);
 function MainController ($scope) {
    if(!window.CONFIG) return;
 
-
-
    $scope.pages = CONFIG.pages;
    $scope.TYPES = TYPES;
    $scope.FEATURES = FEATURES;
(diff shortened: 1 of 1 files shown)


rchl:
* You should use key=value pair for that functionality (for example pageid=foo). People might be currently using some url params already for some functionality. At least I'm using one already for specifying mobile browser.

* the `page.id` key is something you have introduced right? Should add documentation for it.

* If you are calling the key `id` then why comparing it with variable named `desiredPageName`? Doesn't seem to make sense.

* You should update URL state on changing pages. I think it makes sense to remember last set page rather than have it apply for ever after page was loaded.

scripts/controllers/main.js
```diff
@@ -3,6 +3,8 @@ App.controller('Main', ['$scope', MainController]);
 function MainController ($scope) {
    if(!window.CONFIG) return;
 
+
+
```
rchl: Avoid random changes like that. :)

dlashua:
Modified to use pageid= query string parameter.

Parsed QueryString the "hard way" in order to avoid relatively new URLSearchParams class, just in case this is not desirable. 

Modified internal variable name from desiredPageName to desiredPageID

page.id was not introduced by me. It is referenced in the event section in regard to setting up a home assistant event to call $scope.openPage().

Updating the URL to record the last page being viewed could be useful if one presses "reload" or attempts to bookmark a page or share a URL. However, my limited knowledge of javascript in browsers seems to remember that the only way to do this without forcing an actual page reload (which is not needed) would be to use an anchor (#pageid) which TileBoard doesn't seem to like for some reason (?).

rchl:
> page.id was not introduced by me. It is referenced in the event section in regard to setting up a home assistant event to call $scope.openPage().

OK. I see it mentioned in the documentation but as far as I can see, TileBoard doesn't set the id, one has to add it manually to his/her config. It would be good to document this new functionality briefly I think.

> Updating the URL to record the last page being viewed could be useful if one presses "reload" or attempts to bookmark a page or share a URL.

We should think about it and probably address it right away. My use case is a bit different -- on mobile, the browser will evict the page from memory when browser is in the background. That means that TileBoard will reload and revert to initial page. I would like it to stay on last opened page instead.
With your fix it would always revert to initially defined page. It might or might not be expected depending on the use case.
(This problem is even more pronounced when using TileBoard as installed web app. Then it's removed from memory almost immediately after putting it in the background.)

> However, my limited knowledge of javascript in browsers seems to remember that the only way to do this without forcing an actual page reload (which is not needed) would be to use an anchor (#pageid) which TileBoard doesn't seem to like for some reason (?).

It seems that angularjs has some built-in minimal handling that makes that happen. Proper solution could be adding a router (either using angular's ngRoute or external uiRouter), but maybe that's too much.

resoai:
This is just making a mess and is not really relevant. All we need is to add "default" attribute to the page and You can  decide how to mark one as default in the config.

PR #115, opened 2018-08-06
Disable debug console logs by default

rchl:
Just wanted to clean up console which helps especially when developing.
I own a motion sensor that sends state changes for 6 different entities every
few seconds. That makes it easy to miss errors in the console.


resoai:
Thank you!

PR #119, opened 2018-08-13
media_player volume buttons if statement alignment

schmiegelt:
This fixes an issue I have with a silicon frontier internet radio (and probably others). This player supports both VOLUME_SET and VOLUME_STEP. However, the if statement for VOLUME_SET also checks the current level, which this items cannot provide. This means that no volume controls at all are shown. By adding this second condition also to the VOLUME_STEP block ensures that at leasst the plus/minus controls are available.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1004,7 +1004,7 @@
                   </td>
                </tr>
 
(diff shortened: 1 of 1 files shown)


resoai:
This is rather strange. Does HA's UI handle it correctly?

resoai:
I've solved it slightly differently. Could you please check if it works now?

PR #130, opened 2018-08-23
Replace low and high temperature labels with icons

mouth4war:
For your consideration, this PR replaces the High and Low labels for daily temperature with mdi icons (thermometer-lines and thermometer). It matches the icons being used for humidity and wind speed.


resoai:
Thanks but to be frank I don't think that those icons are very obvious

mouth4war:
How about up Arrow and down arrow?

PR #142, opened 2018-09-16
Adding `customStyles` docs

### LINKED ISSUE #141: Missing ability to customise class based on state
Hi,

It's more of a feature request than an issue. 

Currently there is no dedicated option to customise class (or styles) of tile based on its state. 
Potential use case could be air quality monitoring to change background color based on the index or something similar with temperature.

mariuszluciow:
Closes #141


resoai:
This can be both a function or just an object, so I think we need to make it a bit more clear to the users

PR #144, opened 2018-09-17
Changed domain to one used by HASS

mariuszluciow:
UI build in into Home Assistant uses different domains to trigger scripts and switches.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -899,7 +899,7 @@ function MainController ($scope, $location) {
    };
 
    $scope.toggleSwitch = function (item, entity, callback) {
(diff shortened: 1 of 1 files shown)


resoai:
If you look closely to the `toggleSwitch` code you will see that `homeassistant` is the default domain which is than being overridden by others:

```
var domain = "homeassistant";
var group = item.id.split('.')[0];

if(['switch', 'light', 'fan'].indexOf(group) !== -1) domain = group;
``` 

Although this is personal preference, we could update the one calling script. Could you please update the PR?

mariuszluciow:
Oh that true, I missed this bit.

PR #155, opened 2018-09-27
Update index.html

jschwalbe:
Make the config.js load freshly with each refresh by appending the date to the end of config.js (so one doesn't have to clear cache or force reload each time!).

Tested in Apache & Homeassistant (on macOS)

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,7 +32,9 @@
    <script src="scripts/app.js"></script>
 
    <!--<script src="config.example.js"></script>-->
-   <script src="config.js"></script>
+   <!--<script src="config.js"></script>-->
+   <script>document.write("<script src='config.js?v=" + Date.now() + "'><\/script>");</script>
+
 
    <script src="scripts/init.js"></script>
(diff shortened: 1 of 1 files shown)


avishayil:
I think this could be good only for development, while in production you'll probably want browser cache active.

resoai:
Yes, and to be frank in most browsers there is a way to not cache the resource.

jschwalbe:
@resonai @avishayil Hm I respectfully disagree. What is the harm in reloading it each time? Especially when there's so many people trying to figure out why changes don't show up after editing the config.js file?

avishayil:
@jschwalbe Because you're probably running on Raspberry Pi where you want to preserve computing resources. Developers can disable cache on development machines.

resoai:
Embedding javascript files via `document.write` is simply incorrect and causes even more problems. In fact all you need to do is tick one box when you edit the file.

photo_2018-09-28_11-34-24

PR #158, opened 2018-10-07
Added Material Design inspired theme

mariuszluciow:


resoai:
Thank you!

resoai:
We will need to spend some time on the colours as they are a bit all over the place

mariuszluciow:
What do you want to change?

Issue #168, opened 2018-10-26
Vacuum - Return to Home support


mouth4war:
Hi @resoai,

The toggle vacuum tile doesn't support return to home service vacuum.return_to_base. Could you please add support for that?

resoai:
Current vacuum tile acts as a switch only so it does not have many features. I don't have a vacuum myself so even current implementation was done blindly.

mouth4war:
I propose toggling between turn_on and return_to_base. This is how it used to work when using a switch tile before the change in HASS vacuum architecture.

Still think a larger tile based on media_player tile is the way to go eventually. You need no logic, just buttons for play, pause, stop, return to base, locate and spot clean (each button has a service call only). And a fan speed selector to replace the source select.

mouth4war:
For reference, this is how HASS UI vacuum looks while vacuum is cleaning. The toggle button changes to "return to base" from "start cleaning". So yeah, we should toggle between those.

![image](https://user-images.githubusercontent.com/30283635/47600737-95dea480-d9e3-11e8-8390-7b8588cd2017.png)


Detailed view shows all buttons and fan speed selector.

![image](https://user-images.githubusercontent.com/30283635/47600735-81021100-d9e3-11e8-8406-7136f56c8603.png)

resoai:
Not all of the vacuums support "return to base" if I understand correctly. You can see that I already had some logic for various ones.

PR #169, opened 2018-10-27
replace vacuum.stop with vacuum.return_to_base

mouth4war:
Fixes issue #168


mouth4war:
tested and now it works like HASS UI toggle for vacuum.

PR #187, opened 2018-12-24
Added center option for header

FireBall1725:
Ran into a need for wanting to align my home logo in the center on the header. Figured I would share the code incase anyone else had a need for the center area.

image

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -32,6 +32,7 @@ var CONFIG = {
          fontSize: '28px'
       },
       right: [],
+      center: [],
       left: [
          {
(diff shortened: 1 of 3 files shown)


FireBall1725:
Could you elaborate on what it breaks.

resoai:
You are fixing width of the center element to 50% and some folks have left/right elements much longer than that. You have also made changes only to `.css` file and not `.less`.

https://github.com/resoai/TileBoard/wiki/Header-configuration#example

There is an option of defining custom HTML for both left and right while also defining custom css.

PR #190, opened 2019-01-17
- Hide header on to narrow displays, because it overlap the controls

GermanBluefox:
- Add lazy load of the states for ioBroker

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -125,8 +125,12 @@ function MainController ($scope, $location) {
       if(typeof item.id === "object") return item.id;
 
       if(!(item.id in $scope.states)) {
-         warnUnknownItem(item);
-         return null;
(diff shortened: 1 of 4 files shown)


GermanBluefox:
Ping?

Apollon77:
The TileBoard UI is great, I would be happy to see support also for ther smart home systems like ioBroker.
@resoai What do you think? Do you need changes here then please give feedback ...

To join forces especially for UI topics is a great advantage for open source projects

resoai:
Sorry guys, I was travelling with little to no access to the internet. I feel that the correct way would still be to fork the project and do native calls to iobroker rather than trying to map the calls. There is very little still planned to be implemented in TileBoard so syncing any changes would not be too hard. 

There is also a matter of personal opinion which could further slow down the development.

GermanBluefox:
Жаль

Issue #164, opened 2018-10-14
Current Position as Cover state


Nigggo:
Hi there,
I'm just making my first baby steps with TileBoard (and Java Script) and am stuggeling with the covers card.
I'd also +1 for the possibility to show a slider after long press but as an intermediate solution I wanted to show the position as a state instead of just "open" and "close". For some reason the states option does not accept formulas or templates. In the subtitle I could get it to work but I'd prefer to have it up with the state.
Is there a way to solve this or does it require code changes from your side?

```
{
		position: [1, 0],
		type: TYPES.COVER,
		title: 'Esszimmer',
		subtitle: '@attributes.current_position % open',
		id: 'cover.esszimmer',
                 states: {
                       open: 'Open',
                       closed: 'Closed'
                 }
}
```
![image](https://user-images.githubusercontent.com/1877266/46923318-69577f80-d016-11e8-85e7-d3ad7579ca5c.png)

resoai:
Please use state: function

Nigggo:
Thank you. I was 100% sure I tried that before but I must have done something wrong. Now it works :)
What's quite a pitty though is that the tile doesn't seem to fetch the opening position which results in the arrows being incorrectly deactivated if I stop the covers half way (this changes the state to 'open'.
Is there any plan to add full support for covers with position detection?
![image](https://user-images.githubusercontent.com/1877266/46970527-a6c11900-d0b9-11e8-8dd7-9fdeb4b84b1e.png)
![image](https://user-images.githubusercontent.com/1877266/46970679-12a38180-d0ba-11e8-8c59-fcaf394fd7d9.png)

resoai:
We are currently showing arrow controls based on the state (open/closed). Unfortunately I don't have a cover to test it hence the imperfection. Would you be able to show me state dumps at various stages i.e. fully open, fully closed and open half way?

Nigggo:
Of course I will: the state is 'open' for all positions except completely closed as these three screenshots show:
![image](https://user-images.githubusercontent.com/1877266/47039981-78107480-d185-11e8-88b0-fd0f16d6838e.png)

Couldn't you carry over the slider function from the lights card (long press) without much efford? The service to call would be 'cover.set_cover_position' instead of brightness. Or is my lack of coding skills making it look too easy?

Krahen:
Hi, the issue of showing percentage on the state, i've resolved like this:

{
                     position: [0, 0],
                     width: 1,
                     type: TYPES.COVER,
[...]

PR #194, opened 2019-01-30
Fix for partially opened covers #164

domibarton:
By default the cover tile disables the `open` control in case the
entity state is already set to `open`. This might be OK for covers
without positioning details, but not for covers which report their
position back to HA.

Thus, this change will enhance the `open` control to support covers with
partial positions. It will change the behaviour of the `open` control
in the following way. With this change, the `open` control will only be
disabled when:

1. The state is `open` and there's no `current_position` attribute (legacy / backward compatibility)

2. The state is `open` and the `current_position` attribute equals `100` (new position support)

### DIFF
--- a/index.html
+++ b/index.html
@@ -482,7 +482,7 @@
 
          <div class="item-cover">
             <div class="item-cover--button -open"
-                 ng-class="{'-disabled': entity.state === 'open'}"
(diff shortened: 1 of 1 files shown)


resoai:
Thank you!

PR #200, opened 2019-02-08
set secrets and URLs in a separate file

mgafner:
I modified config.example.js and index.html to set secrets in a separate file, so the config.js of a running setup can be easy shared on github

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -6,6 +6,14 @@
  Make sure you use real IDs from your HA entities.
 */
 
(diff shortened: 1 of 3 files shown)


resoai:
In all fairness this creates even more problems for those getting started and I hardly find JSON.parse() justifiable for such cause let alone debugging becomes impossible. I have decent idea about splitting config into multiple files which I’m planning to share once I find a moment and my health condition will allow it.

PR #202, opened 2019-02-14
Disabling text selection

arvathrowaway:
This edit disables text selection on the page, eliminating situations where long presses accidentally select text. To my knowledge, this doesn't mess with any other touch functionality.


Evgeny-:
We use .less for developing, so i changed original .less file with your merge. Thank you.

Issue #203, opened 2019-02-14
no answer after resume


madmicio:
I'm using tilebord on a dedicated tablet. I use kiosk browser, but I have the same problem on other devices and chorme.
when I load the page for the first time, everything works correctly. when the device goes into standby and I resume tilebord, it does not seem to be connected to home assistant. no command works until the page is reloaded.
depends on what?  is There a solution?

resoai:
Make sure tablet does not disconnect WiFi when it goes to standby.

Evgeny-:
Try to update repo and use `reconnectAfterScreensaver: true` in the root of config

madmicio:
in the config.js file like this?
/*
 This is an example configuration file.

 COPY OR RENAME THIS FILE TO config.js.

 Make sure you use real IDs from your HA entities.
*/
   reconnectAfterScreensaver: true,
var CONFIG = {
   customTheme: CUSTOM_THEMES.TRANSPARENT etc.etc.etc.

Evgeny-:
@madmicio 

```
var CONFIG = {
   reconnectAfterScreensaver: true,
   customTheme: CUSTOM_THEMES.TRANSPARENT,
.......
```

rchl:
I have the same issue sometimes on my phone. TileBoard installed on desktop as an "app". After resume I get one of those behaviors:
- "System error" notification
- Invalid token notification (can't remember exact message)
- Just empty app background with no entities

In most of those cases I get error notification and then entities appear after short while. In rare cases waiting doesn't seem to help and I need to reload page.

I have disabled screensaver so this change (11f626cdcf7267235081e67d80e3160e2f9270f1) wouldn't even have a chance to help. But looking at this change, it seems like quite a hack. Is there specifying reasoning behind it? It seems quite weird to try to reconnect on *showing screensaver*. Is the idea that this code should trigger after resume because then all queued actions would trigger that didn't have a chance to run during sleeping? But why not do it on hiding screensaver then?

Anyways, maybe it would be useful to just analyze with logs what exactly happens (websocket connection status and such) on all of those cases and try to find more specific solution? I could do job myself if you are interested.

Evgeny-:
@rchl i probably get this issue wrong. For some reason i thought about screensaver. So i did incorrect fix. Tbh i don't know how to solve your common problem. May be we should try to use some analog of ping.

madmicio:
[...]

PR #207, opened 2019-03-13
Improve ping and reconnect logic

apop880:
#203 should be resolved by this. I've cleaned up the logic around pinging to use the standard ping and pong options provided by the Home Assistant WebSocket interface, and I added some additional checks to ensure that pings don't continue to be sent if the connection is stopping or starting. I've thoroughly tested and this has resolved reconnection issues I was having.


scripts/controllers/main.js
```diff
@@ -1931,20 +1931,21 @@ function MainController ($scope, $location) {
...
 
-      Api.getUser(function (res) {
+      Api.sendPing(function (res) {
          if('id' in res) success = true;
       });
 
       setTimeout(function () {
          if(success) return;
 
+         realReadyState = false;
```
Evgeny-: Why does it needed?

README.md
```diff
@@ -67,6 +67,12 @@ var CONFIG = {
     */
    authToken: null,
 
+   /* pingMaxTimeout: Set to false disable pinging of the websocket connection.
+    * Otherwise, a ping will be sent every five seconds, and if a response is not received in 3 seconds,
+    * a reconnect will be attempted. If not included in the config file, setting defaults to true.
+    */
+   pingMaxTimeout: true,
```
Evgeny-: Seems a little bit obsolete name, since it's not about "max timeout"

Evgeny-:
Thank you for that. I've added few comments on questionable lines, can you please comment them?

Evgeny-:
Thank you for that. Should we change `pingMaxTimeout` name?

README.md
apop880: I updated ```pingMaxTimeout``` to ```pingConnection```, I agree that makes more sense.

apop880:
Pushed a new commit with ```pingMaxTimeout``` changed to ```pingConnection``` and an existing reference to ```pingMaxTimeout``` that I didn't catch before removed from the readme. Let me know if you have any other questions.

Evgeny-:
Everything seems good. Thank you

PR #208, opened 2019-03-18
Add support for armed_night state

Juggels:
Home Assistant supports the `armed_night` state. This PR adds support for this state to the Alarm card.

### DIFF
--- a/index.html
+++ b/index.html
@@ -180,14 +180,22 @@
 
             <div class="alarm-popup-buttons-line">
                <div class="alarm-popup-button -icon -home"
+                    ng-if="['disarmed'].includes(alarmEntity.state)" 
                     ng-click="actionAlarm('alarm_arm_home', activeAlarm, alarmEntity)">
                   <div class="mdi mdi-bell-plus"></div> Arm home
                </div>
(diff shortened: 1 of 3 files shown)


Evgeny-:
We had troubles with .includes, #63 people use tileboard on IE. I think will be better create separate function in main controller (something like `isDisarmed`?) and use .indexOf instead.

resoai:
Isn't armed_home identical to armed_night ?

Juggels:
> We had troubles with .includes, #63 people use tileboard on IE. I think will be better create separate function in main controller (something like `isDisarmed`?) and use .indexOf instead.

Good one, I did not think to check that. I will refactor accordingly.

Juggels:
> Isn't armed_home identical to armed_night ?

That depends I guess... They are implemented separately in Home Assistant. For my personal use case, they differ in use as well: `armed_home` is used for an extra sense/layer of security when someone is home alone, whereas `armed_night` is used (as the name suspects) to trigger on some other stuff as well during the night time.

Evgeny-:
@Juggels thank you. I've moved changes from .css file to .less (we use .less for developing, .css is basically translation)

Issue #209, opened 2019-03-25
Not able to use a SLIDER for cover


alexbarcelo:
I may be doing something wrong. But I was trying to have a cover controlled by a `TYPES.SLIDER` and could not get the slider bar to have correct information.

I have:

```
{
  position: [1, 0],
  id: 'cover.cover_living',
  title: 'MyCover',
  type: TYPES.SLIDER,
  unit: '%',
  state: false,
  value: '@attributes.current_position',
  slider: {
    request: {
      type: "call_service",
      domain: "cover",
      service: "set_cover_position",
      field: "position"
    }
  }
},
```

What am I missing? I tried to add a `value` in the top level dictionary, or add a `field` item with no luck.

I believe that [this line](https://github.com/resoai/TileBoard/blob/6119c17ae307c67c1adcc45b1a7f29feed51ee5f/scripts/controllers/main.js#L717) may be the key point, but... I don't fully understand the variables. And I have run out of ideas.

alexbarcelo:
To add a little more info on what I tried:

This shows an arbitrary position:

```
...
slider: {
  value: 35,
  ...
}
...
```

But I tried things like `value: '@attributes.current_position'` and does not work, maybe the context is not valid? I also tried to put full `&cover.cover_living.attributes.current_position'` with no luck, same Angular error:

`Error: "[ngModel:numfmt] http://errors.angularjs.org/1.7.2/ngModel/numfmt?p0=%40attributes.current_position"`

alexbarcelo:
Solved by #210 

Thanks @Evgeny- !

I am not a js guy, and I was fearing being very wrong from the start :)

PR #210, opened 2019-03-25
Use `field` mechanism for value in sliders

alexbarcelo:
Just like the LightSliderConf, allow the configuration to include a `field` and use that as the attribute name to be used for the value of the slider.

This is the same the is being done in `getLightSliderConf`, so seems suitable in that context, if I understood everything correctly. This was my approach to solve #209, but I may have overlooked "the proper way", maybe?

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -709,12 +709,13 @@ function MainController ($scope, $location) {
 
       var def = item.slider || {};
       var attrs = entity.attributes || {};
-
+      var value = +attrs[def.field] || 0;
+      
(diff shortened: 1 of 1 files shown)


Evgeny-:
> but I may have overlooked "the proper way", maybe
Probably you didn't. Seems no one actually tried this type of tile, or at least no one among developers

Evgeny-:
(But yes, we have a way to parse expressions, may be there are better way to add parser to the tile `state` property, e.g. `var value = +parseFieldValue(def.value, item, entity) || entity.state || 0`)

PR #216, opened 2019-04-03
Ignore any *config files in site root

danhounshell:
So you can have additional config files, like mobile-config.js, ipad.config.js, etc. at the site root and have them be ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 
 /sensors.js
 /config.js
+/*config.js
 .idea/*
 .idea
 *.map


Evgeny-:
There is "/includes/" folder in the gitignore for such proposes

danhounshell:
I understand and I make use of the includes folder after finding that in the .gitignore file. Though if I have multiple config files I would like to have them in the same place in the file structure. But I understand if not wanting to accept this change as there is already a place, the includes folder, to add custom code.

resoai:
I believe that having a single config.js and whatever user wants in `/includes` is a much cleaner solution.

Issue #128, opened 2018-08-22
Add RGB light color picker


mouth4war:
Would it be possible to integrate a color picker for color selection of RGB lights?

resoai:
This would require breaking light secondary controls into a popup (like alarm and date picker) and a lot of changes to the way light tile behaves. Unfortunately I don't have time to do this.

mouth4war:
May I propose using a new button for color picker in addition to the +/- buttons for brightness. This one can be on the left as a special button only for RGB lights. This way the sliders remain as is and color picker can be added easier.

I believe there are existing libraries for color picking and these return a value that can be passed to HASS.

resoai:
There is simply no space for it there really. It would probably be much simpler to make a slider which would control hue alone of HSL, assuming Saturation is fixed at 100% and Lightness is basically a value of our brightness.

mouth4war:
That would be a good workaround thanks. Can you give me an example so I can try?

resoai:
This still requires some work to be done in the code to make that happen. Sadly I absolutely have no time for it at the moment.

cgarwood:
I think it would be great to be able to specify configuration for a tile to have a more info popup if you long-press it (similar to the more-info dialogs in the default HA UI), but that would be a pretty large undertaking to add in.

TheCellMC:
I totally agree with @cgarwood. I think it would be very cool if we could have a colour picker on long press. Maybe somethings like the colour picker of the TileBoard project? 
![screen shot 2018-08-30 at 09 28 57](https://user-images.githubusercontent.com/12169936/44836711-fe094800-ac37-11e8-812a-a198c4f71ffc.png)

isbrodie1:
Hey guys, has there been any further update on this one?

MPopti0n:
As a work around (no as nice as a color picker...)

1. **I created scripts in `HomeAssistant` in `script.yaml`** (one for each color)

```
color_hue_livingroom_green:
  alias: "Couleur hue - Salon - VertFonce"
  sequence:
    - service: mqtt.publish
      data:
        topic: "Maison/lumieres/in"
        payload : "HA-Couleur-Salon-VertFonce"
        retain: false
```
_I use Node-Red/MQTT but you can use direct instructions to the entity in HA depending on your light (mine are Philips Hue) like this :_

 ```
color_hue_livingroom_green:
  alias: "Couleur hue - Salon - VertFonce"
  sequence:
    - service: light.turn_on
      data:
[...]

PR #221, opened 2019-04-09
Added RGB Color picker

danhounshell:
Added RGB Color picker since it is a popular request: #128 

Based on: https://github.com/Alberplz/angular-colorpicker-directive with a few slight modifications to get it working well with TileBoard (I probably ought to fork that repo and keep track of the changes that I made)

Color picker is available on the 2nd view of light tile if the entity has an `rgb_color` attribute.

Only the RGB properties are sent to HA, the Alpha property, if set, is discarded.

Screenshot of it in action:

RGBColorPickerTileBoard


danhounshell:
Could probably use a little more tweaking to get rid of the alpha controls altogether but I ran out of time for the day and I wanted to get this posted before I got sidetracked onto something else.

danhounshell:
Commented out some of the functionality from color-picker that we are not using. Now looks like this:

RGBColorPickerTileBoard_2

resoai:
Thank you for your effort! I'm sure a lot of users will appreciate it. 

Perhaps if you will have some free time in future it would be nice to get rid of "R" "G" and "B" if you are considering forking it and styling cancel button.

Evgeny-:
Thank you for commit. Please in the next time respect our code style, and don't forget that we use es4

danhounshell:
@Evgeny- I saw your follow up commit and noted the things you updated. Thank you.

PR #222, opened 2019-04-10
Changed color picker to be config option rather than implicit

danhounshell:
Changed color picker to be a config option for lights rather than
implicitly diplayed if entity has rgb_color attribute (though they will
not do anything in HA if entity does not have rgb_color attribute).
Fixed bug so that color picker can be used when sliders are not being
used. Fixed margins to better match sliders.

if light config has the following then color picker will be displayed on the longpress view:
```
colorpicker: true,
```
Perhaps future changes will have the colorpicker configuration to be more like sliders and allow you to specify which attribute to update, which type of color values to use (RBG, HEX, or HLSA), etc. I only have one color changing bulb to test with, others may be different.

### DIFF
--- a/index.html
+++ b/index.html
@@ -682,7 +682,7 @@
                </div>
             </div>
           
-            <div ng-if="entity.attributes.rgb_color" class="item-entity-colorpicker">
+            <div ng-if="item.colorpicker" class="item-entity-colorpicker">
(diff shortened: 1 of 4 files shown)


danhounshell:
Also, once this is merged in I will update the docs and the tile examples file.

PR #238, opened 2019-06-16
Adde Mapbox map type

jaydensmith:
I added Mapbox as an available option for the device tracker tile, as it has high free usage limits and better options.


resoai:
Thank you!

PR #244, opened 2019-07-15
feat: support swiping to switch pages

### LINKED ISSUE #242: [Feature request] Swipe to toggle page
On mobile it would be natural to allow switching pages using swipe action instead of trying to hit tiny page selector buttons.

I have actually implemented this for myself but changes are fairly extensive so not sure if you would be interested. I did this:
 - using hammer library (also replaces currently used 'long.press' library)
 - switches page stacking to be horizontal when menu is at the bottom (for swipe direction to follow toggle animation)

rchl:
Use hammer library to add support for swiping to switch to next/previous
page.

When menu is on the left, changing pages is done using swipe up/down
motion and when on the bottom, using swipe left/right motion.

With menu on the bottom, for page toggle animation to match swipe
direction, it was necessary to change how pages are layouted (needed to
be horizontally instead of vertically).

angular-long-press library was replaced with hammer as it supports the
same and was needed for swiping anyway.

Fixes #242


resoai:
Thank you! Could you please also include .less with changes made to CSS?

resoai:
Perfect! 

P.S. I have also added you as a maintainer :)

PR #249, opened 2019-08-04
Fully qualified entity_picture url

fahad-hasan:
I use direct links for facebook profile pictures and entity picture and the tracks keeps appending the server url to it. Small fix working locally.


resoai:
Thank you

PR #250, opened 2019-08-25
Fixed cover tile wrongly disable open function.

emilp333:
Covers report "open" as long as they are not fully closed. Only then the state changes to "closed". The cover tile would disable the open function until the cover was fully closed. If you have covers down at 50% you would need to get down to closed before you could open them again. 
This change looks at the "current_position" attribute and only disable the open function if the position is 100%, meaning fully open. 
If the cover does not support position then the open function will never be disabled. 
This is the same behaviour as Home assistant.

### DIFF
--- a/index.html
+++ b/index.html
@@ -482,7 +482,7 @@
 
          <div class="item-cover">
             <div class="item-cover--button -open"
-                 ng-class="{'-disabled': entity.state === 'open'}"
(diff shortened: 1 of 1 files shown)


emilp333:
I now realised that I did not have the latest code. In fact it was almost a year old and this has already been fixed. Sorry for that. I'll close this one :)

PR #251, opened 2019-08-28
Fix item and entity not being passed to getWeatherLine

inxaos:
This was making writing anonymous methods for the line section of the weather module much more difficult.


resoai:
Thanks!

Issue #233, opened 2019-05-23
[BUG] config.serverURL added before camera URL


JesseWebDotCom:
I'm trying to bypass home assistant (192.168.1.50) and get a live MJPG stream from camera (192.168.1.25).  If I use the code below:
```
{
   position: [0, 8],
   width: 5,
   height: 3,
   type: TYPES.CAMERA,
   id: {},
   bgSize: 'contain',
   state: false,
   fullscreen: {
      type: TYPES.CAMERA,
      bgSize: 'contain',
      filter: function (item, entity) {
         return "http://192.168.1.25:81/mjpg/front_door/video.mjpg";
      }   
   },
   filter: function (item, entity) {
      return "http://192.168.1.25:81/mjpg/front_door/video.mjpg";
   }
}
```

Tileboard will suffix my filter URL with config.serverURL like this:
```
background-image: url(http://192.168.1.50:8123http://192.168.1.25:81/mjpg/front_door/video.mjpg?_i=0);
```

How can I fix this?
Thanks

JesseWebDotCom:
I fixed this by updating directives.js - I replaced this line:
```
 if(url) el.style.backgroundImage = 'url(' + CONFIG.serverUrl + url + ')';
```

with this:
```
            if(url) 
            {
               if(url.toLowerCase().startsWith("http")) {
                  el.style.backgroundImage = 'url(' + url + ')';
               } else {
                  el.style.backgroundImage = 'url(' + CONFIG.serverUrl + url + ')';
               }
            }   
```

This fix ensures only relative URL's are prefixed by CONFIG.serverUrl

PR #254, opened 2019-09-05
Added check for several BG images to determine if they are local or not

inxaos:
This resolves https://github.com/resoai/TileBoard/issues/233 by adding checks in several places to see if the URL is relative or not

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -240,6 +240,14 @@ function MainController ($scope, $location) {
       return true;
    };
 
+   function setBackgroundImage(image){
+      if(image.toLowerCase().startsWith("http")) {
+         return 'url(' + image + ')';
+      } else {
+         return 'url(' + CONFIG.serverUrl + image + ')';
+      }
+   }
+
    $scope.pageStyles = function (page, index) {
       if(!page.styles) {
          var styles = {};
@@ -253,11 +261,7 @@ function MainController ($scope, $location) {
             var sbg = parseFieldValue(page.bgSuffix, page, {});
 
             if(sbg) {
-               if(sbg.toLowerCase().startsWith("http")) {
-                  styles.backgroundImage = 'url(' + sbg + ')';
-               } else {
-                  styles.backgroundImage = 'url(' + CONFIG.serverUrl + sbg + ')';
-               }
+               styles.backgroundImage = setBackgroundImage(sbg);
             }
          }
 
@@ -426,11 +430,7 @@ function MainController ($scope, $location) {
 
             if(bg) 
             {
-               if(bg.toLowerCase().startsWith("http")) {
-                  styles.backgroundImage = 'url(' + bg + ')';
-               } else {
-                  styles.backgroundImage = 'url(' + CONFIG.serverUrl + bg + ')';
-               }
+               styles.backgroundImage = setBackgroundImage(bg);
             }
          }
 
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -252,7 +252,14 @@ function MainController ($scope, $location) {
          else if(page.bgSuffix) {
             var sbg = parseFieldValue(page.bgSuffix, page, {});
 
-            if(sbg) styles.backgroundImage = 'url(' + CONFIG.serverUrl + sbg + ')';
+            if(sbg) 
+            {
+               if(sbg.toLowerCase().startsWith("http")) {
+                  styles.backgroundImage = 'url(' + sbg + ')';
+               } else {
+                  styles.backgroundImage = 'url(' + CONFIG.serverUrl + sbg + ')';
+               }
+            }
          }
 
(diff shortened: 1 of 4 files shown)


scripts/controllers/main.js
```diff
@@ -252,7 +252,14 @@ function MainController ($scope, $location) {
          else if(page.bgSuffix) {
             var sbg = parseFieldValue(page.bgSuffix, page, {});
 
-            if(sbg) styles.backgroundImage = 'url(' + CONFIG.serverUrl + sbg + ')';
+            if(sbg) 
+            {
+               if(sbg.toLowerCase().startsWith("http")) {
```
rchl: - Would make sense to have a helper function now that the code is bigger and needed in 3 places. Could call it `toAbsoluteURL`
 - `startsWith` is fairly new feature of ecmascript. Probably around same time as let/const and this project doesn't use those either. I'm not original author of the code so maybe you shouldn't listen to me but just making you aware of possible constrains.

scripts/controllers/main.js
```diff
@@ -252,7 +252,14 @@ function MainController ($scope, $location) {
          else if(page.bgSuffix) {
             var sbg = parseFieldValue(page.bgSuffix, page, {});
 
-            if(sbg) styles.backgroundImage = 'url(' + CONFIG.serverUrl + sbg + ')';
+            if(sbg) 
+            {
```
rchl: brace style doesn't match this project's style

scripts/controllers/main.js
inxaos: Internet Explorer doesn't seem to support "StartsWith" but it seems like most other stuff going back the last 7 or 8 years does, based on what I'm seeing here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith.

I see your point about creating a helper method for the check, I'll give that some thought.

scripts/controllers/main.js
inxaos: fixed

scripts/controllers/main.js
```diff
@@ -240,6 +240,14 @@ function MainController ($scope, $location) {
       return true;
    };
 
+   function setBackgroundImage(image){
```
rchl: `GetBackgroundImageUrl` please as it doesn't set anything.

scripts/controllers/main.js
```diff
@@ -240,6 +240,14 @@ function MainController ($scope, $location) {
...
+      if(image.toLowerCase().startsWith("http")) {
```
rchl: Check the code in `$scope.trackerBg`. It's also same stuff that should be unified. And for safety, please use `indexOf` instead of `startsWith`, like that code.

scripts/controllers/main.js
```diff
@@ -240,6 +240,14 @@ function MainController ($scope, $location) {
...
+         return 'url(' + image + ')';
+      } else {
+         return 'url(' + CONFIG.serverUrl + image + ')';
```
rchl: Style: This code uses two spaces for indentation

scripts/controllers/main.js
```diff
@@ -240,19 +240,27 @@ function MainController ($scope, $location) {
       return true;
    };
 
+   function toAbsoluteURL(image) {
+      if(image.indexOf('http') !== 0) {
```
rchl: That condition seems backwards. serverUrl won't be added if `http` is missing at the beginning.

scripts/controllers/main.js
```diff
@@ -240,19 +240,27 @@ function MainController ($scope, $location) {
...
+
    $scope.pageStyles = function (page, index) {
       if(!page.styles) {
          var styles = {};
 
          if(page.bg) {
             var bg = parseFieldValue(page.bg, page, {});
 
-            if(bg) styles.backgroundImage = 'url(' + bg + ')';
+            if(bg) styles.backgroundImage = toAbsoluteURL(bg);
```
rchl: This actually breaks loading of page background in my particular case.
That's because `bg` is `images/bg3.jpg` so doesn't have leading `/` and `serverUrl` doesn't have trailing slash. So concatenating produces wrong URL.

Probably `toAbsoluteURL` should check that and add slash if needed between joined parts.

scripts/directives.js
```diff
@@ -31,7 +31,10 @@ App.directive('camera', function () {
          var appendImage = function (url) {
             var el = document.createElement('div');
 
-            if(url) el.style.backgroundImage = 'url(' + CONFIG.serverUrl + url + ')';
+            if(url)
+            {
+               el.style.backgroundImage = toAbsoluteURL(url);
```
rchl: In this file/scope, there is no access to `toAbsoluteURL` function so code will throw.

inxaos: Fixed

rchl:
I've fixed the remaining issues and generalized the function so that it can be used in two more places.

Feel free to review my commit @inxaos.

PR #255, opened 2019-09-05
Added climate mode fix from HA forum user athua

inxaos:
Fixes climate mode control which was broken by HA climate changes in v0.96 .

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1273,10 +1273,10 @@ function MainController ($scope, $location) {
       sendItemData(item, {
          type: "call_service",
          domain: "climate",
-         service: "set_preset_mode",
+         service: "set_hvac_mode",
          service_data: {
             entity_id: item.id,
-            preset_mode: option
+            hvac_mode: option
          }
       });
 
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1270,15 +1270,27 @@ function MainController ($scope, $location) {
       $event.preventDefault();
       $event.stopPropagation();
 
-      sendItemData(item, {
-         type: "call_service",
-         domain: "climate",
-         service: "set_hvac_mode",
-         service_data: {
-            entity_id: item.id,
-            hvac_mode: option
-         }
-      });
+      if (entity.attributes.preset_mode){
+         sendItemData(item, {
+            type: "call_service",
+            domain: "climate",
+            service: "set_preset_mode",
+            service_data: {
+               entity_id: item.id,
+               preset_mode: option
+            }
+         });
+      } else {
+         sendItemData(item, {
+            type: "call_service",
+            domain: "climate",
+            service: "set_hvac_mode",
+            service_data: {
+               entity_id: item.id,
+               hvac_mode: option
+            }
+         });
+      }
 
       $scope.closeActiveSelect();
 
--- a/index.html
+++ b/index.html
@@ -1011,17 +1011,17 @@
                <span ng-if="(_unit = entityUnit(item, entity))"
                      class="item-climate--target--unit" ng-bind="_unit"></span>
             </div>
-            <div class="item-climate--mode" ng-if="entity.attributes.preset_mode"
-                 ng-click="openSelect(item)">
-               <span ng-bind="entity.attributes.preset_mode"></span>
+            <div class="item-climate--mode" ng-if="entity.attributes.hvac_modes"
+                 ng-click="openSelect(item)" ng-touch="openSelect(item)">
+               <span ng-bind="entity.state"></span>
             </div>
          </div>
 
          <div ng-if="selectOpened(item)" class="item-select"
-              ng-style="itemSelectStyles(entity, entity.attributes.operation_list)">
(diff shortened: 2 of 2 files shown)


bdf0506:
I disagree with this PR as is. There needs to be more logic added, as it only takes into account dumbed down thermostats that have functions of Heat/Cool/Auto/etc and nothing more.

For example, my thermostat has the logic of Activities. Therefore, the way TileBoard interacts with my thermostat as-is without any TileBoard changes, I have the ability to simply select Away, Sleep, Wake, etc from the dropdown list. These map on the backend to PRESET_HOME, PRESET_SLEEP, etc.

Sure, I can't select from my dropdown of TileBoard to say Heat/Cool/Auto/etc (which maps to HVAC_HEAT, HVAC_COOL, etc), but the above activities are far more useful, since the activity can actually define the heat/cool/auto action.

inxaos:
@bdf0506  can you provide config samples of how your thermostat is configured?  I'd be happy to take a look and try to figure out how to make this work in both cases

bdf0506:
What do you want exactly? I am using this component for my thermostat, which is a bit complex, sits it has to sit on top of another open source component. But I would assume there are other native components that handle this similar to how mine does.
https://github.com/MizterB/homeassistant-infinitude

inxaos:
I think this should support both the original way and also now work if you only have hvac modes.  I set it so that it won't display the hvac modes if you have preset modes, but I can add a config option or something if you think we need to support both at the same time.

resoai:
Nope. This ruins climate tile for those who have presets. Suppose we need to specify which parameter to use i.e. hvac mode or preset

scripts/controllers/main.js
```diff
@@ -1270,15 +1270,27 @@ function MainController ($scope, $location) {
...
-         service_data: {
-            entity_id: item.id,
-            preset_mode: option
-         }
-      });
+      if (entity.attributes.preset_mode){
+         sendItemData(item, {
+            type: "call_service",
+            domain: "climate",
+            service: "set_preset_mode",
```
rchl: To avoid duplicating almost the same code twice, you could use a ternary expression:

```js
service: entity.attributes.preset_mode ? "set_preset_mode" : "set_hvac_mode",

Comment (resoai) [reply scripts/controllers/main.js]: I have KNX room controllers and I have both preset and HVAC mode so my understanding was that all functionality is now being shifted to preset mode. One possible workaround for those affected could be to use preset mode by default and add an attribute hvacMode: true or something similar in tile definition.

Comment (resoai) [reply scripts/controllers/main.js]: It would appear that I have hvac_modes: auto (list with one mode) but not the actual hvac_mode.

```
hvac_modes: auto
current_temperature: 24
min_temp: 15
max_temp: 27
target_temp_step: 0.5
temperature: 21
preset_mode: comfort
preset_modes: comfort,away,sleep,eco
friendly_name: AV room
supported_features: 17
```

**Comment by chadcb** [comment]
> I think this should support both the original way and also now work if you only have hvac modes. I set it so that it won't display the hvac modes if you have preset modes, but I can add a config option or something if you think we need to support both at the same time.

This would be a great option to have. I have a Nest and it uses both modes (preset/hvac). I notice the Home Assistant climate card is using the hvac_modes and hvac_action as the status. So I prefer the hvac option based on how we use it and to mimic HA.

Something like this in the CLIMATE tile config?

```
// Use hvac_modes and hvac_action instead of preset_modes and preset_mode
hvacMode: true
```

It may also be useful to mimic HA in displaying the status of both if using hvac. Example, if the heat is on it shows "Heating" in the HA Climate Card (Thermostat). If I change preset_mode to "Away", it shows "Heating - Away". If preset mode is none, it just shows "Heating".

Thanks!

Here are the climate attributes that show up in HA for the Nest:
```
hvac_modes: auto,heat,cool,off
current_temperature: 71
min_temp: 50
max_temp: 90
temperature: 71
target_temp_high: null
target_temp_low: null
fan_mode: auto
fan_modes: on,auto
hvac_action: idle
preset_mode: none
preset_modes: none,away,eco,Away and Eco
friendly_name: Dining Room
supported_features: 27
```

**Comment by christophebaraer** [comment]
I would concur with @chadcb's comment and the need to both support HVAC and Preset modes. 
```
// Use hvac_modes and hvac_action instead of preset_modes and preset_mode
hvacMode: true
```
seems like a good idea to me also to "switch" between those two modes. I do have a Zwave underfloor heating thermostat and it only supports hvac mode.

mouth4war:
HVAC mode is the main way of controlling climate and the tile is broken for this reason. Preset mode is an additional way of controlling via Presets. 

Can we at least merge the HVAC mode fix to make this tile usable again?

scripts/controllers/main.js
mouth4war: Hey @resoai Check for hvac_action, if available. That shows the current action for selected mode (eg. if mode is heat, action can be heating or idle).

rchl:
I don't have any way to test this stuff but if you feel confident please prepare a **minimal** change that fixes that one thing and hopefully we can get it in.

resoai:
This should now be fixed in #384

PR #256, opened 2019-09-06
Implemented ng-gauge as a tile option.

inxaos:
This implements a customized version of angular-gauge as a tile option.  This should address issue 253.

Readme and Tile examples have been updated with the new tile options.


helenclarko:
Fantastic, very excited about this!

helenclarko:
Makes me happy that you also updated the Tiles Example md file.

scripts/controllers/main.js
```diff
@@ -899,6 +899,32 @@ function MainController ($scope, $location) {
           && entity.state !== 'off';
    };
 
+   $scope.getGaugeField = function (field, item, entity) {
+
+      let def = { bg_color: 'rgba(0, 0, 0, 0.1)',
```
rchl: This project doesn't use `let`, `const` or any fancy ES futures.

scripts/controllers/main.js
```diff
@@ -899,6 +899,32 @@ function MainController ($scope, $location) {
...
+                  size: 200, 
+                  duration: 1500,
+                  thick: 6,
+                  type: 'full',
+                  min: 0,
+                  max: 100,
+                  cap: 'butt',
+               };
+
+      if(!item || (!item.settings[field] && !def[field])) return null;
```
rchl: This logic crashes if item doesn't have `settings` object defined. It doesn't get to use the defaults in that case.

scripts/controllers/main.js
rchl: Can you use meaningful name for this variables - `defaults` instead of `def`.

scripts/controllers/main.js
```diff
@@ -899,6 +899,32 @@ function MainController ($scope, $location) {
           && entity.state !== 'off';
    };
 
+   $scope.getGaugeField = function (field, item, entity) {
+
```
rchl: nit: unnecessary empty line

scripts/controllers/main.js
rchl: Also, using `!` is not good way of checking if setting is defined as setting might be defined but evaluate to false and that should be OK. You can use `field in item.settings` instead.

scripts/controllers/main.js
```diff
@@ -899,6 +899,32 @@ function MainController ($scope, $location) {
...
+      let def = { bg_color: 'rgba(0, 0, 0, 0.1)', 
+                  fg_color: 'rgba(0, 150, 136, 1)', 
+                  size: 200,
```
rchl: It would be nice if size would by default auto-adjust to tile size. For example the logic could maybe be: 80% of tile width or height (whichever is smaller).

resoai:
Please also make sure you include changes to .less file in the PR.

styles/main.less
```diff
@@ -2086,6 +2086,10 @@ camera-thumbnail {
    }
 }
 
+.gauge{
```
rchl: I think according to convention it should be `.item-gauge`. And could find a better place for it as it's in the middle of some irrelevant styles.

inxaos:
> There are various styling/spacing/indentation issues that I won't bother reporting. If I'm the one merging, I'll just fix everything up.

How are you checking these?  I'm typically a C# backend programmer, so I'm fairly unfamiliar with the JS ecosystem for this kind of stuff.

README.md
```diff
@@ -408,6 +408,25 @@ Tile Object. [Click here for some real life examples](TILE_EXAMPLES.md)
...
+              thick: 8, //Defaults to 6
+              label: 'My Gauge', //Defaults to undefined
+              append: '@attributes.unit_of_measurement', //Defaults to undefined                           
+              prepend: '$', //Defaults to undefined
+              duration: 1500, //Defaults to 1500ms
+              label_only: false, //Defaults to false
+              fg_color: 'rgba(0, 150, 136, 1)', //Defaults to rgba(0, 150, 136, 1)
+              bg_color: 'rgba(0, 0, 0, 0.1)', //Defaults to rgba(0, 0, 0, 0.1)
+              fraction_size: 0 //Number of decimal places to round the number to. Defaults to current locale formatting
+             }
```
rchl: - Can you also change indentation to two spaces here?
- And add space after comment start (`//`)
- add a trailing `,` after the object end. Otherwise it's not a valid javascript (even though it's just readme)

scripts/controllers/main.js
```diff
@@ -899,6 +899,36 @@ function MainController ($scope, $location) {
           && entity.state !== 'off';
    };
 
+   $scope.getGaugeField = function (field, item, entity) {
+      var gauge_defaults = {
+         bg_color: 'rgba(0, 0, 0, 0.1)',
+         fg_color: 'rgba(0, 150, 136, 1)',
+         size: function () { return .5 * (CONFIG.tileSize * (item.height > item.width ? item.height : item.width)); },
```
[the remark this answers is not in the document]
inxaos: I'll do some testing and update this accordingly

jdestefanis:
H! I took your not published code and I have added the parameter: thresholds
Let me know if you want me to push the changes.
Best!

inxaos:
> H! I took your not published code and I have added the parameter: thresholds
> Let me know if you want me to push the changes.
> Best!

Go right ahead

helenclarko:
@inxaos

rchl:
- I've changed options to be in `camelCase`
 - Added support for `thresholds` option
 - Taking smaller of width/height to make gauge fit in rectangular tiles
 - Fixed some style

rchl:
So it's good from my side, second pair of eyes would be good.

resoai:
Looks good! Thanks!

PR #258, opened 2019-09-09
fix: wrong initial month in INPUT_DATETIME picker

### LINKED ISSUE #235: TYPES.INPUT_DATETIME defaults to wrong month
When I click on TYPES.INPUT_DATETIME the input box that comes up shows the previous month - ie currently shows 05 when it should show 06 for June.

rchl:
Since `getMonth()` returns 0-based month, we need to add 1 to get
current month.

Resolves #235


resoai:
Thank you!

PR #264, opened 2019-10-05
Add automated configuration functions and some examples

akloeckner:
As I got confused by all the different braces and did not like to repeatedly define the same options for similar tiles, I set up some auto-configuration functions. I am sure, they will need some more fine-tuning. So, this PR is to start the discussion about it. Right now, I have changed nothing in the original code, but only added things to `config.js`.

What is new here?
* `TEMPLATES` defines functions for common options to repeatedly used tiles.
* `autoTile()` uses these functions automatically based on the entity `id`.
* `autoGroup()` arranges those tiles automatically to a group.
* `importGroup()` imports a group from the HA states.
* `importPage()` imports a whole page from the HA states (group with `view=true`).

The functionality is split up in four files:
* `config.js` only imports the remaining three files.
* `templates.js` defines the `TEMPLATES`.
* `autoconfig.js` defines the functions explained above.
* `variable.js` is the new `config.js`. I couldn't think of a better name...

Finally, in `variable.js` there are snippets, showing how to 
* use the templates
* use the `auto*` functions (may be done on init of TileBoard)
* use the `import*` functions (must be done after the `HApi` is available).

### DIFF
--- a/autoconfig.js
+++ b/autoconfig.js
@@ -7,11 +7,13 @@
  */
 function autoTile(id, position, args) {
    // Match the id against the templates (sorted to match most specific first)
-   var template = Object.keys(TEMPLATES.FILTERS).sort().reverse().find(
-      function(k) {
-         return id.match(k) !== null
-      }
-   );
+   var template;
+   if (typeof id != 'undefined')
+      template = Object.keys(TEMPLATES.FILTERS).sort().reverse().find(
+         function(k) {
+            return id.match(k) !== null
+         }
+      );
    // If none is found, use the default
    if (typeof template === 'undefined')
       template = 'default';
--- a/autoconfig.js
+++ b/autoconfig.js
@@ -1,0 +1,160 @@
+
+/**
+ * AUTOMATIC TILES
+ * 
+ * Generates a tile object from (id, position, args) tuple.
+ * Uses the logic as explained above for TEMPLATES-FILTERS.
+ */
+function autoTile(id, position, args) {
+   // Match the id against the templates (sorted to match most specific first)
+   var template = Object.keys(TEMPLATES.FILTERS).sort().reverse().find(
+      function(k) {
+         return id.match(k) !== null
+      }
+   );
+   // If none is found, use the default
+   if (typeof template === 'undefined')
+      template = 'default';
+   // If states are available and id is not found there, use 'missing' template
+   if (typeof(angular.element(document.body).scope()) !== 'undefined'
+    && typeof(angular.element(document.body).scope().states[id]) === 'undefined')
+      template = 'missing';
+   // Return the TEMPLATE's resulting tile object
+   return TEMPLATES.FILTERS[template](id, position, args);
+}
+
+/**
+ * AUTOMATIC GROUPS
+ * 
+ * This function takes a group object with an array of tile stubs and 
+ * generates a proper "items" list with proper "tile" objects from it.
+ * The stubs should be arranged in a matrix as follows:
+ *   [
+ *     ['sun.sun', , 'sensor.uptime'   ]
+ *     [         , , 'group.all_lights']
+ *   ]
+ * Each stub can be a simple entity_id. Then, the TEMPLATES above are used.
+ * If the stub is a dictionary, its "id" element is used as "id" and the rest is merged.
+ * The "position" element is generated autmatically, but can be (uselessly) overriden by
+ * defining it in the stub.
+ *
+ * Each row of tiles is wrapped, in order to respect the group's "width" property,
+ * if provided. As a side effect, you can provide only one long row of tile stubs and
+ * the tiles will be arranged to match the group's "width" property.
+ *
+ * Take care to provide a double array (i.e. [[...]]) as items!
+ */
+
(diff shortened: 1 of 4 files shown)


akloeckner:
So, after using TileBoard for a while with this setup, I would suggest to move the functions and the minimal templates to `app.js`. Keeping the custom templates and the actual configuration back in `config.js`.

Together with #282, I would then even go as far as suggesting to incorporate the default `action` in the templates (maybe renamed to `DEFAULTS`). By applying `DEFAULTS` in the `ng-repeat` directive for each tile, this would allow for customizing a tile type very easily.

Next step could then be to enable/disable the parts of a generic super-tile with options in the defaults, e.g. `hasPlusButton: true` and `plusAction: ...` for `TYPES.LIGHT`.

akloeckner:
Ah, and I found that I am not using the `import*` functions at all. I just found it hard to do everything I wanted with HA groups and the templates only. So, this part could be removed. Or kept, someone else might have a use for it after all...

akloeckner:
Sorry, mis-clicked...

rchl:
IMO, this is a sort of change that opens a whole another world of bugs/feature requests/tweaks that people would request. Basically I feel like this should basically be a separate project. It's really hard to justify including something opinionated like that in this project.

The ideal solution would be to have a plugin API where someone could plug-in this behavior via a plugin. That would also get rid of all the hacks in this code for getting access to scopes and such.

Sorry for being skeptical, I appreciate your effort. I have also created some custom code to auto-align tiles as fiddling with sizes and positions is not fun but I don't currently see a nice way of integrating those.

resoai:
In my config I tend to keep definition of tiles separately thus my definition of a group looks like this:

```
function $$(tileObject, merge) {
   if(merge instanceof Array) {
      merge = {position: merge};
   }

   var copy = mergeObjects({}, tileObject);

   return mergeObjects(copy, merge);
}
```
```
{
    title: 'Outside',
    width: 2,
    height: 3,
    items: [
        $$(TILES.DOOR_ENTRY_FRONT_GATE, [0, 0]),
        $$(TILES.OUTSIDE_LIGHTS, [1, 0]),
        $$(TILES.FRONT_GATE, [0, 1]),
        $$(TILES.REAR_GATE, [1, 1])
    ]
}
```

This also allows me to define tiles dinamically:

```
 {
         title: 'Music',
         bg: 'images/bg2.png',
         icon: 'mdi-music',
         groups: [
            {
               title: 'Zones',
               width: 3,
               height: 3,
               items: []
                  .concat([
                     "lounge",
                     "dining_room",
                     "kitchen",
                     "bedroom",
                     "kids_room",
                     "study",
                     "guest_room",
                  ].map(function (key, i) {
                     var id = 'media_player.' + key;

                     var y = Math.floor(i / 3);

                     return {
                        position: [i - y * 3, y],
                        id: id,
                        type: TYPES.MEDIA_PLAYER,
                        hideMuteButton: false,
                        state: false,
                     };
                  }))
            },
```

There is a multitiude of ways to make your config look readable and while being rather hacky it gives you a lot of freedom.

akloeckner:
Your `$$` looks quite like my `TEMPLATES`. So, that would let me think, it might be good to introduce something like this. 

But the “multitude of ways“ makes me think, that it might actually be a bit bold to impose one specific way to organise their config on the user.

Should we maybe
- just start a section in the wiki for everyone to showcase their ways of organizing? 
- and add some way to include more than one custom js file? Like @rchl's plugin idea...

resoai:
My config.js looks like this:

```
function loadScript(url) {
  var req = new XMLHttpRequest();
  req.open('GET', url + '?cache=' + Math.random(), false);
  req.onreadystatechange = function(){
     if (req.readyState === 4) {
        var s = document.createElement('script');
        s.appendChild(document.createTextNode(req.responseText));
        document.head.appendChild(s);
     }
  };
  req.send(null);
}

var conf = location.search.match(/config=([\w\d_-]+)/);
conf = conf ? conf[1] : 'default'
loadScript('includes/' + conf + '.js');
```

This is by no means corect way of loading javascript files synchronosly but it works for me :) Showcasing ways to do configs in wiki is a good idea.

akloeckner:
I believe there is no (much) more correct way of loading JS synchronously. I did some web research now and came up with this snippet, which works fine:
```
function loadJS(url) {
   var xhttp = new XMLHttpRequest();
   var script = document.createElement( "script" );
   xhttp.open("GET", url+suffix, false);
   xhttp.send();
   script.text = xhttp.responseText;
   document.head.appendChild( script ).parentNode.removeChild( script );
}
```
I took the "evaluation" part from `jquery`, so this seems to be "standard".
https://github.com/jquery/jquery/blob/cf84696fd1d7fe314a11492606529b5a658ee9e3/src/core/DOMEval.js#L39

Obviously, using jquery it becomes a one-liner, such as here:
https://stackoverflow.com/a/28010938

It can be used like this:
```
loadJS('templates.js'); //defines var TEMPLATES
alert(TEMPLATES);
```

So, until we introduce plugins or whatever, just documenting such a function in the wiki would help in splitting the configuration.

PR #267, opened 2019-10-10
Allow Add-To-Homescreen feature

akloeckner:
This commit adds a manifest.json to allow adding TileBoard to the cellphone's homescreen as a 'WebApp'.

See details here:
https://developer.mozilla.org/de/docs/Web/Manifest


manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "TileBoard",
+  "short_name": "TileBoard",
+  "start_url": "./index.html",
```
rchl: I have an issue with this because some (me, for example), might want to start with a custom URL. In my case I start with `?mobile=1` param. If there is no manifest, or if this key is not set, I'm able to do that because browser will use current URL when adding app/shortcut. With this key being here, the browser will use this URL rather than URL at the time of creating the app/shortcut.

manifest.json
```diff
@@ -0,0 +1,13 @@
...
+  "display": "fullscreen",
```
rchl: Fullscreen hides every phone UI element, including the top status bar. It might be good for some usecases, but personally I'd prefer `standalone` which hides browser UI but leaves phone status and navigation keys. It's probably a matter of personal way of using this app. Some might run TileBoard exclusively on a tablet in which case `fullscreen` is nice, but some might just occasionally open it on a device like phone and in that case removing whole phone UI is a bit too much.

Personally I'm also using `orientation = portrait` since landscape wouldn't work correctly with my layout but for some either might be fine.

So I think we're never gonna please everyone and thus I'm not sure it's a good idea to force any particular settings in manifest. I'm not really sure what's the way forward here.

resoai:
We could add an example to Wiki but I agree @rchl that we should not force users, especially since this changes the behaviour people are used to.

akloeckner:
I believe, it should be as easy as possible fit the average user to get a symbol on their home screen. That would mean, a minimum manifest should be there by default, because the option did only appear to my Firefox mobile when I added the file. It could then be part of the wiki to explain how to customize the file then.

I'll check this and add another commit...

As a reference, this is what HA itself generates:
```
{"background_color": "#FFFFFF", "description": "Home automation platform that puts local control and privacy first.", "dir": "ltr", "display": "standalone", "icons": [{"sizes": "192x192", "src": "/static/icons/favicon-192x192.png", "type": "image/png"}, {"sizes": "384x384", "src": "/static/icons/favicon-384x384.png", "type": "image/png"}, {"sizes": "512x512", "src": "/static/icons/favicon-512x512.png", "type": "image/png"}, {"sizes": "1024x1024", "src": "/static/icons/favicon-1024x1024.png", "type": "image/png"}], "lang": "en-US", "name": "Home Assistant", "short_name": "Assistant", "start_url": "/?homescreen=1", "theme_color": "#03A9F4"}
```

index.html
```diff
@@ -12,10 +12,14 @@
 
    <title>TileBoard</title>
 
+   <!-- WebApp configuration -->
+   <link rel="manifest" href="./manifest.json" crossorigin="use-credentials">
```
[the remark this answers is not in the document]
akloeckner: I also thought so, until I read this:
https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes

> The use-credentials value must be used when fetching a manifest that requires credentials, even if the file is from the same origin.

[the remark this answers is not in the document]
resoai: Does it really need `./` in the url?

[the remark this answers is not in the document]
akloeckner: No, I checked and removed the `./`

manifest.json
akloeckner: Good point, I removed that member.

manifest.json
akloeckner: I removed that as well. And left only necessary members in the file.

akloeckner:
So, in the latest commit, I removed all un-neccessary members from the manifest JSON. It appears, name (Firefox uses the URL otherwise) and icon (Firefox doesn't offer "add to home" otherwise) are the only ones required.

PR #269, opened 2019-10-12
fix spelling and grammer errors

REVATHYJESS:


rchl:
Old text wasn't removed so now things are duplicated.

resoai:
It is also really hard to see exactly what changes have been made to text :)

rchl:
I've re-created this PR at #278 with stuff correctly changed. :) Thanks.

PR #270, opened 2019-10-14
Code reformatting

cadavre:
Hi, I have a dozen of changes that I can contribute to the project:
* charts
* HLS camera streaming
* more tiles (icon, weather_icon, service, sensor_chart, camera_stream, door_entry_stream)
* better error handling
* additional features
* icon notifications on screensaver
* more "modes" for styling in different environments (desktop, tablet, mobile)
* and more...

But one thing that keeps me from PRs is poor code formatting in JS and HTML. This commit unifies all JS code formatting.

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -6,278 +6,279 @@
  Make sure you use real IDs from your HA entities.
 */
 
-
 var CONFIG = {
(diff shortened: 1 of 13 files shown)


rchl:
Sorry, but this will likely never go through. Style for this project is different and it doesn't make sense to force anyone's personal style because then every other person would try to change it. I think you would need to rather just follow the project's style.

What makes sense is creating an eslint configuration in this project to enforce rules that would follow the current style. Unfortunately this project is not an npm project so it makes things difficult in that respect. Otherwise it would be easy to create lint script and commit hooks that would guard proper style.

rchl:
Also, it's generally not a good idea to reformat vendor scripts.

cadavre:
True, vendors shall not be reformatted.

PR #271, opened 2019-10-17
Enable/Disable Swipe

obimichael:
Implemented to turn of swipe in config. If attribute swipeEnabled is missing in config, swipeEnabled = true.

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -27,7 +27,7 @@ var CONFIG = {
    menuPosition: MENU_POSITIONS.LEFT, // or BOTTOM
    hideScrollbar: false, // horizontal scrollbar
    groupsAlign: GROUP_ALIGNS.HORIZONTALLY, // or VERTICALLY
-   swipeEnabled: true, // enable swipe between Pages (default: true)
+   enableSwipe: true, // enable swipe between Pages (default: true)
 
    header: { // https://github.com/resoai/TileBoard/wiki/Header-configuration
       styles: {
--- a/config.example.js
+++ b/config.example.js
@@ -27,6 +27,7 @@ var CONFIG = {
    menuPosition: MENU_POSITIONS.LEFT, // or BOTTOM
    hideScrollbar: false, // horizontal scrollbar
    groupsAlign: GROUP_ALIGNS.HORIZONTALLY, // or VERTICALLY
+   swipeEnabled: true, // enable swipe between Pages (default: true)
 
    header: { // https://github.com/resoai/TileBoard/wiki/Header-configuration
       styles: {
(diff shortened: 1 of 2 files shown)


resoai:
In such a case, this should probably be disableSwipe and you need to use “===“ operator to avoid type conversion and there is no need to check for null.

obimichael:
First of all sorry for the many commits, my first time on github. ;-)
You're right and I changed the operator, but I would leave "enabled" in the parameter name, because of the other parameter wording

rchl:
I wonder if it should have more explicit naming. `enableSwipe` refers to swiping action and not to the feature that it enables/disables. I think it's a bit too low-level. In a hypothetical situation where we would use swipe for more actions, would we want those to also be disabled? I think potentially not. So I think better naming could be `disablePageToggleWithSwipe`.

rchl:
Also, any particular reason for disabling it?
If it conflicts with some workflow of yours, maybe that could be improved/fixed instead?
Adding options adds complexity and potential for bugs in the feature.

akloeckner:
Maybe this is the reason? 

https://community.home-assistant.io/t/tileboard-new-dashboard-for-homeassistant/57173/2702

rchl:
I saw that but then it would make sense to either fix that or exclude older safari versions automatically.

akloeckner:
Oh no, please let's not do that. I'm very happy that things just work on my stone age iPad...

rchl:
By excluding I meant to disable the feature that causes the problem. In this case swiping to toggle pages. :)

akloeckner:
Ouff. ;-) 

Btw. The post in the forums is with respect to an iPhone on iOS 13.x. So, it's not an old version.

resoai:
> 
> 
> Maybe this is the reason?
> 
> https://community.home-assistant.io/t/tileboard-new-dashboard-for-homeassistant/57173/2702

I think that this was my worry when we introduced `Hammer.js`.

rchl:
> I think that this was my worry when we introduced Hammer.js.

I'll have a look when I have some time.

But I'm not sure what you mean by "this" in "this was my worry" :)
The issue is not necessarily with the hammer library itself. It could be that we are not preventing some event. Or could be that just having a touch handler higher in the three conflicts with native range handler, irrespective of the library type.

resoai:
I mean that I tried adding swipe gestures at the very beginning and faced some issues with other controls, so I decided to abandon the idea :)

rchl:
Created #302 to fix issues with sliders.

resoai:
@obimichael belive we can close this one now?

obimichael:
okay

PR #274, opened 2019-10-19
Add a history chart popup to sensors

akloeckner:
uses Chart.js 2.8.0
uses Moment.js 2.24.0
(both MIT license)

Things to consider:
* I basically started with a copy of the iframe-popup. 
* Should the History-Retrieval be moved to the Api? 
* Should it be more general? HA itself is much more involved about its history charts.
* I'm not using some fetch Api or similar on purpose: this should work on older systems, too.
* ~~How can I add the 'tick' in the top-left corner of the tile to indicate it can be clicked? ~~
  Figured that one out: add `classes` to your tile as below.
* This adds two more dependencies. 

Here is an example tile with all options:
```

{type: TYPES.SENSOR, id:
                  'sensor.temperatur_aussen_gefiltert',  
classes: ['-switch'],
historyClasses: 'item -th-cover',
  historyStyles: {
    color: 'red',width: '80%'
  },
  historyOffset: 24*60*60*2*1000,
  historyEntity: 'sensor.temperatur_innen_gefiltert',
  historyChart: {
    type: 'scatter',
data:{datasets: [{borderColor: 'rgb(255,0,0)'}]}
  }
},
```


scripts/controllers/main.js
```diff
@@ -1499,6 +1502,85 @@ function MainController ($scope, $location) {
...
+      for (k in styles) popupHistoryStyles[k] = styles[k];
+
+      return popupHistoryStyles;
+   };
+
+   $scope.openPopupHistory = function (item, entity) {
+      $scope.activeHistory = item;
+
+      // Use REST API to get history
+      var xhttp = new XMLHttpRequest();
```
rchl: Angularjs has built-in API for making request - https://docs.angularjs.org/api/ng/service/$http

scripts/controllers/main.js
```diff
@@ -1499,6 +1502,85 @@ function MainController ($scope, $location) {
...
+      var startDate = new Date( Date.now()
+                              - ($scope.itemField('historyOffset', $scope.activeHistory, entity) || 24*60*60*1000)
+                              ).toISOString();
+      var entity_id = $scope.itemField('historyEntity', $scope.activeHistory, entity) || entity.entity_id;
+      xhttp.open("GET", CONFIG.serverUrl+'/api/history/period/'+startDate+'?filter_entity_id='+entity_id, false);
```
rchl: - Since you've asked, yes I think this should be part of `HApi`
- This doesn't seem to work due to missing authorization
- Is there no way to get that data through websocket connection?

styles/main.css
```diff
@@ -1021,8 +1025,9 @@ camera-thumbnail div {
 }
 .camera-popup-container,
 .door-entry-popup-container,
-.iframe-popup-container {
-  background: #111;
+.iframe-popup-container,
+.history-popup-container {
+  background: #111;border: 1px white;
```
rchl: Should change the original scss file and compile it.

scripts/controllers/main.js
akloeckner: I think, I did try a construct with `.then`, but it didn't work on my old iPad, which is my main target for TileBoard. But I'll check out `$http` in the course of the above comments...

scripts/controllers/main.js
akloeckner: * Ok, I'd agree.
* I was surprised, but it seems to work through some magic cookie or the likes.
* I looked hard and didn't find a websocket method for that. Even the lovelace interface seems to use REST for the history. I'd like to be proven wrong, because that would simplify things.

styles/main.css
akloeckner: Ah! And I already wondered, how you guys maintain this huge style sheet... I'll need help with that (first scss contact for me).

akloeckner:
Let's wait for the high-level ok and then go forward with reviews, changes, documentation...

styles/main.css
rchl: There is a quickstart on the front page here http://lesscss.org/

scripts/controllers/main.js
rchl: It should be compatible as long as angularjs is - https://docs.angularjs.org/misc/faq#what-browsers-does-angularjs-work-with-

scripts/controllers/main.js
rchl: > I was surprised, but it seems to work through some magic cookie or the likes.

It doesn't seem to work for me. Maybe you have whitelisted access to API for your IP or you have some cookie saved that I don't. You could test second theory by testing in incognito window.

rchl: See https://developers.home-assistant.io/docs/en/auth_api.html#making-authenticated-requests
You need to include `Authorization` header.

scripts/controllers/main.js
akloeckner: I believe, Safari for iOS 9.3.5 (my old iPad) is not one of
> the latest versions of Chrome, Firefox, Safari, and Safari for iOS

:-)

I just found, the original TIleBoard code also uses the old API. Maybe for a reason...
https://github.com/resoai/TileBoard/blob/d8f02e43b7c027f87383e9ba4e12be9149c11194/scripts/models/api.js#L264

I'll still try quickly, when I get around to it...

resoai:
I don’t mind history in a pop-up in principal however the implementation is a bit sketchy and ads even more dependencies. All transport functions should definitely be in API.

scripts/controllers/main.js
akloeckner: Interesting. It works also in private mode. And I only have a `trusted_networks` in my HA configuration... But if the docs say, we need to athenticate, I'll try to add that when moving stuff to the API. 

Can you advice how to use the token system in the API? I just read across it and I am not sure, I fully understand how to use it.

akloeckner:
The dependency could be brought down (seemingly) by moment.js, because there is a version, where it is bundled to chart.js. But the code would still be there obviously. This is why I included both separately. It seemed cleaner. But yes: It adds two dependencies.

And the sketchy code can be improved with some guidance. I'm not a JS developer orginally...

rchl:
One problem for me with introducing this feature is that it does only cover some use cases. History of entity would be useful for most if not all entities, not only sensors. It is nice being able to see when switches (lights) were turned on or off, for example. Or how long TV was running. I understand that the problem is that some entities already react to click so there is a conflict but maybe we should think of more radical changes then, that would implement graphs globally in a consistent way.

(Easier said than done though...)

scripts/controllers/main.js
rchl: You just add Authorization header to the request:
```
Authorization: Bearer ABCDEFGH
```
Where `ABCDEFGH` you can find at `$Api._token` currently, I believe.

akloeckner:
How about a more flexible configuration for click and long press? We could allow a field like `action` or `longPressAction` in all tile types, thus enable overriding default click actions. If someone wants history, they could just put `openPopupHistory(item, entity)` in that option. 

Another possibility would be to introduce a small symbol like the `-clickable` tick, which can be actually clicked. As I don't see another 'free' place, this should probably replace the clickable tick. It could become a small square, if a popup is associated to the tile.

cgarwood:
I would love each tile to have an overridable onclick or onlongpress function.

akloeckner:
In #282 I found that actually only the dimmer tile has a longClickAction associated. So, the longClickAction would be a perfect candidate to show a history graph, if the user wishes so.

rchl:
> So, the longClickAction would be a perfect candidate to show a history graph, if the user wishes so.

I would really wish we'd come up with a solution that is more user-friendly than having to manually specify handlers when wanting to show graph for an entity. As I said, it really makes sense to support graph for all entities.

I feel like default would have to be "long press" but then there would need to be a solution for how to make that coexist with existing functionality (light dimming).

I have one suggestion but it would require significant changes -- instead of changing state of tile on long press, have a "sheet" slide up from the bottom with relevant options. That would have enough space to accommodate both graphs and tile-specific options. Or it could have some sort of tabs interface.

akloeckner:
That's basically already there: popups. I'm pretty sure there is a straightforward css option to make it slide.

Then: How about to create a generic popup? Much like the camera one, where one could add tiles as desired:
```
{
  ... tile stuff ...,
  popup: {
    type: fullscreen/ontile,
    width: ..., height: ...,
    items: [
      {
        type: TYPES.HISTORY,
        id: 'some template to get id of tile'
      },
      {
        type: TYPES.SLIDER, ...
      }
    ]
  }
}
```

Finally, associate some default `popup` configuration to tile types. Much as my `TEMPLATES` in the other PR. ;-) But let's not mix this up for now.

styles/main.css
akloeckner: so, I did that. much smaller change now.

however: there are new useless cssrules now. I guess that is just an acceptable side-effect?

also: a few things turn out to compile differently than in master brach. e.g. leading zeros and a few comments. i haven't found options to reproduce the ouput as it was before.

scripts/controllers/main.js
akloeckner: It's working for me, because my HA is so “old“...

```
Access from trusted networks without auth token is going to be removed in Home Assistant 0.96. Configure the trusted networks auth provider or use long-lived access tokens to access /api/history/period/2019-11-08T09:24:35.626Z from ...
```

rchl:
I will do a separate PR for turning API into a module. It's needed to be able to use $http inside of it.

akloeckner:
Wouldn't it make sense to keep the API independent of the app? So that people could reuse it? 

Also, if you change it, there is already a potential use for `$http` in the current code. I'm a bit afraid that changing that bit of code will break things on older devices, which do not support promises. Such as my old iPad on ios 9.3.5.

rchl:
@akloeckner 
I've actually just done that. See:
03073990905a0b84dea224779e2b5900fef63b7f
d86d7f8cbd44af181a8146359d32d3d89a0e92af
b58a8546d4d6a46bba601fe9883be7170f592b0d

To answer your questions:
> Wouldn't it make sense to keep the API independent of the app? So that people could reuse it?

Being an angular service, it can still be used independently by injecting it wherever anyone wants.
But then I'm not sure what do you mean by `reuse`? Do you mean reuse within TileBoard or completely independently? I think the latter case is not the goal and we shouldn't need to care about that.

> Also, if you change it, there is already a potential use for $http in the current code.

Yes, I've converted existing use of `XMLHttpRequest` into `$http`.

> I'm a bit afraid that changing that bit of code will break things on older devices, which do not support promises

Angular.js includes its own implementation of Promises. It doesn't depend on native promises. If angular.js itself works on a device, there is no reason to think that it's internal `$http` service won't.

akloeckner:
By reuse, I meant actually re-use it anywhere. But you're right: not our primary concern.

I checked now on my old device and `$http` works just fine. (Pew...)

rchl:
A couple more comments after looking at the code more:
 - Can use https://jtblin.github.io/angular-chart.js/ instead of messy manual use of Chartjs API.
 - Graphs only work for some values. Works for temperature for example but not for binary sensors with only on and off value.
 - The graph is quite unreadable IMO. Compare the graph for the same sensor with how HomeAssistant shows it:

Home Assistant | This PR
---------------- | --------
 | 
 - Since we can easily make the graph smaller, there would be plenty of space to move light dimming to the popup and then just enable long press to show the graph for all entities. Like Home Assistant does it in fact:

styles/main.css
```diff
@@ -579,7 +579,7 @@ body {
 .item.-device_tracker .item-title,
 .item.-camera .item-title,
 .item.-camera_thumbnail .item-title {
-  text-shadow: 1px 1px 0 rgba(0, 0, 0, 0.2);
+  text-shadow: /*1px 1px 1px rgba(0,0,0,.2),*/ 1px 1px 0 rgba(0, 0, 0, 0.2);
```
rchl: Hmm?

styles/main.css
```diff
@@ -352,7 +352,7 @@ body {
   opacity: 0;
   position: absolute;
   top: 0;
-  transition: opacity 0.3s;
```
rchl: For me leading zeros are not removed after running: `lessc main.less main.css`

styles/main.less
```diff
@@ -1206,17 +1206,17 @@ camera-thumbnail {
...
       display: inline-block;
       vertical-align: top;
       background: #222;
+   }
 
-      > iframe {
-         height: 100%;
-         width: 100%;
-      }
+   &--iframe > iframe, &--canvas > canvas {
```
rchl: Why did you not just modify existing, nested rule? Like `> iframe, > canvas { ... }`?

styles/main.css
akloeckner: That's the output from `lessc`... The comment is in the `main.less` already...

styles/main.less
akloeckner: Because the `iframe` only is in the `--iframe`. As is the `canvas` only in the `--canvas`. Had I added only `> iframe, > canvas`, even more useless CSS rules would have been created.

styles/main.css
akloeckner: Do we have different `lessc` versions maybe? 

```
lessc -v
lessc 1.6.3 (LESS Compiler) [JavaScript]
```

Ok. Quick check told me: That's very old. But that's what I get with `sudo apt-get install node-less`...

rchl: npm packages are best installed with npm.
As it says on its page, you should use `npm i -g less`. And you need `node` and `npm` of course.

akloeckner: I fear, I am lost here. `apt-get` does not have `node` nor `npm`. It has `nodejs` and that is already installed. But on the `npm` site it says, `npm` is installed with `node`. Both `node -v` nor `npm -v` are unknown commands on my pi. 

So, last resort, I went for an online converter, e.g. https://www.webtoolkitonline.com/less-to-css.html. But the results vary widely (in newlines and indentation mainly), which clutters the commit even more. Also, the leading zeros are still missing and the strange comment is still in the `css`.

I am out of ideas how to make my setup reproduce your `css` code.

rchl: Node includes `npm` since a long time. Maybe your version of node is very old.
You just have to google how to install latest node (or at least v10) on your system.
Since you said you are using RaspberryPi, I think this should work: https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions

akloeckner: Thanks! That worked indeed.

styles/main.css
akloeckner: should be ok now

scripts/controllers/main.js
```diff
@@ -1499,6 +1500,100 @@ function MainController ($scope, $location) {
...
+      var startDate = new Date( Date.now()
+                              - ($scope.itemField('historyOffset', $scope.activeHistory, entity) || 24*60*60*1000)
+                              ).toISOString();
+      var entity_id = $scope.itemField('historyEntity', $scope.activeHistory, entity) || entity.entity_id;
+
+      $http({
+         method: 'GET',
+         url: CONFIG.serverUrl+'/api/history/period/'+startDate+'?filter_entity_id='+entity_id,
+         headers: {
+            "Content-type": "application/json"
```
rchl: There is no need to set Content-Type as you are not sending any data (this is the *request* data type).

scripts/controllers/main.js
```diff
@@ -1499,6 +1500,100 @@ function MainController ($scope, $location) {
...
+      updateView();
+
+      var startDate = new Date( Date.now()
+                              - ($scope.itemField('historyOffset', $scope.activeHistory, entity) || 24*60*60*1000)
+                              ).toISOString();
+      var entity_id = $scope.itemField('historyEntity', $scope.activeHistory, entity) || entity.entity_id;
+
+      $http({
+         method: 'GET',
+         url: CONFIG.serverUrl+'/api/history/period/'+startDate+'?filter_entity_id='+entity_id,
```
rchl: Use `toAbsoluteServerURL` instead of prefixing manually with `CONFIG.serverUrl`

scripts/controllers/main.js
```diff
@@ -1499,6 +1500,100 @@ function MainController ($scope, $location) {
...
+   };
+
+   $scope.openPopupHistory = function (item, entity) {
+      $scope.activeHistory = item;
+      updateView();
+
+      var startDate = new Date( Date.now()
+                              - ($scope.itemField('historyOffset', $scope.activeHistory, entity) || 24*60*60*1000)
+                              ).toISOString();
+      var entity_id = $scope.itemField('historyEntity', $scope.activeHistory, entity) || entity.entity_id;
```
rchl: style: variable should be called `entityId`.

scripts/controllers/main.js
resoai: Also, please always add spases around `+`

resoai:
I would agree with @rchl and it would be much cleaner to stick with `angular-chart.js`

scripts/controllers/main.js
rchl: And of course, this code should still be moved to api.js. You would need to rebase your changes on master.

styles/main.css
resoai: Try running `npm update` which should update all packages.

resoai:
Also, just a thought: wouldn't it look cleaner like this:

```
history: {
  offset: 24*60*60*2*1000,
  entity: 'sensor.temperatur_innen_gefiltert',
  chart: {
    type: 'scatter',
    ...
  }
}
```

akloeckner:
Alright. I think, I addressed everything I can for now:

Structural things:
- use `$http`. done
- move things to `Api`. done
- modify `.less` and compile. done
- check if this is possible through websocket. done. it is not possible.
- dependencies. we will have to live with it, if we want this feature
- angular-chartjs. this is not maintained anymore. also it would introduce even more dependencies. for completeness: a new option might be ng2-charts.
- sketchy code. better now, I hope. please advise.
- move options to lower-level `history`. done
- consistent showing of popup. done. it is now default on long-press for sensors. and everything else will show it, if `history` is present in the tile object.
- new options documented. done

Minor things
- add authorization. done
- no content-type. done
- spaces around `+`. done, i hope.
- use `toAbsoluteServerURL`. done
- variables should be CamelCased. done

Additional features requested
- categorial data. such as binary sensors. done
- chart too big. it just fills the popup. if data is pretty depends on the data actually. and the device orientation. any maybe other things.
- unreadable graphs. I just used the minimum options i could think of. let's talk about default graph options now. 

Additional features requested but of scope for this PR
- generic sliding options opopup. this is out of the scope of this PR, I believe.
- default chart types. this is out of scope of this PR, I believe.

styles/main.css
```diff
@@ -1069,13 +1075,19 @@ camera-thumbnail div {
...
   height: 100%;
   width: 100%;
 }
 .camera-popup--iframe,
 .door-entry-popup--iframe,
-.iframe-popup--iframe {
+.iframe-popup--iframe,
+.history-popup--iframe,
+.camera-popup--canvas,
+.door-entry-popup--canvas,
```
resoai: Canvas should only be for `.history-popup-X`

index.html
```diff
@@ -46,6 +46,8 @@
          'scripts/vendors/angular.hammer.js',
          'scripts/vendors/color-picker.js',
          'scripts/vendors/angularjs-gauge.js',
+         'scripts/vendors/moment.min.js', // Required by Chart.js for time axes
```
resoai: Since we are including `moment.js` we shuld probably use it for `timeAgo` function which is now in `app.js`. This is outside of scope of this PR but is probably worth to remember.

rchl:
> angular-chartjs. this is not maintained anymore. also it would introduce even more dependencies. for completeness: a new option might be ng2-charts.

Many angular.js components are not maintained anymore. Even angular.js is basically on life support. I don't think we should be worried about that as long as the component works fine right now.

Sorry, but I'd like to see it done properly using that directive. The way things in angular should be done and what makes reactivity work nicely without needing to manually draw or detect when things change.

ng2-charts is for angular2+, not for angular.js.

rchl:
> chart too big. it just fills the popup. if data is pretty depends on the data actually. and the device orientation. any maybe other things.

I think it makes sense to force certain ratio. For example 12:6 (haven't tried so up to discussion), so that graph is always wider than higher. The width needs to be as big as possible as it's most important to have time axis to be as precise as possible. The unit axis probably doesn't need a lot of space typically and can even make things less readable.

styles/main.css
akloeckner: Check

akloeckner:
Aspect ratio is 4:3 now including title and axes, which looks good to me. That might even roughly fit your 2:1 on the actual chart area.

akloeckner:
I also invested quite some time on the angular module. I'm still not convinced, but it's your project.

scripts/models/api.js
```diff
@@ -111,6 +111,31 @@ App.provider('Api', function () {
          return this.socket.send(wsData);
       };
 
+      $Api.prototype.rest = function (request) {
+         var r = request;
```
rchl: - Still overriding original object here as `r` is a reference to `request`.
 - Please don't use meaningless variable names like `r`. Can call function argument 'requestOptions' and this one `request`, for example.

scripts/models/api.js
```diff
@@ -111,6 +111,31 @@ App.provider('Api', function () {
...
+               if (response.status >= 400 && response.status <= 499) {  // authentication error
+                  redirectOAuth();
+               } else {
+                  return null;
+               }
+            });
+      };
+
+      $Api.prototype.getHistory = function (startDate, filterEntityId) {
+         var r = {
```
rchl: Please use a meaningful variable name.

scripts/models/api.js
```diff
@@ -111,6 +111,31 @@ App.provider('Api', function () {
...
+            type: 'GET',
+            url: '/api/history/period/'
+         };
+         if (startDate) r.url += startDate;
+         if (filterEntityId) r.url += '?filter_entity_id=' + filterEntityId;
```
rchl: This code's logic gives impression that both arguments are optional and I don't think they are. The API will:
 - error if startDate is not provided
 - if filterEntityId is not provided, it will return history for all entities (I believe) but only draw first (?)

And since the error in the request will trigger re-authentication, current logic should take that into account. You could triger error notification if required arguments are not provided.

scripts/models/api.js
```diff
@@ -111,6 +111,31 @@ App.provider('Api', function () {
...
       };
 
+      $Api.prototype.rest = function (request) {
+         var r = request;
+         r.url = toAbsoluteServerURL(r.url);
+         r.headers = r.headers || {};
+         r.headers.Authorization = 'Bearer ' + this._token;
+         return $http(r)
+            .catch(function (response) {
+               if (response.status >= 400 && response.status <= 499) {  // authentication error
```
rchl: I did that mistake in another place but not all those codes are authentication errors (404 is "not-found", for example) :)
So if this API returns 404 (which it does for some parameters/entities), it will trigger authentication dialog.

scripts/models/api.js
resoai: > 
> 
> Now it looks a bit bad due to lacking any padding within the popup, don't you think?:
> graph-spacing
> 
> Also, there is no legend/entity name. Did you drop that on purpose?

Padding needs to be added and perhaps a mobile theme could have a smaller close button?

scripts/models/api.js
```diff
@@ -111,6 +111,31 @@ App.provider('Api', function () {
          return this.socket.send(wsData);
       };
 
+      $Api.prototype.rest = function (request) {
+         var r = request;
+         r.url = toAbsoluteServerURL(r.url);
+         r.headers = r.headers || {};
+         r.headers.Authorization = 'Bearer ' + this._token;
+         return $http(r)
+            .catch(function (response) {
```
rchl: I'd add:
```js
.then(function (response) {
   return response.data;
}
```
before `catch` so that we only return data from this api. Returning whole response object shouldn't be necessary.

scripts/controllers/main.js
```diff
@@ -1492,6 +1497,65 @@ App.controller('Main', ['$scope', '$location', 'Api', function ($scope, $locatio
...
+               numbers.push(stateInfo.state);
+            });
+            labels.push(Date.now());
+            numbers.push($scope.states[entityId].state);
+
+            // Populate chart data
+            $scope.activeHistory.labels = labels;
+            $scope.activeHistory.data = [numbers];
+            $scope.activeHistory.series = [$scope.states[entityId].attributes.unit_of_measurement ? ($scope.states[entityId].attributes.friendly_name + ' /' + $scope.states[entityId].attributes.unit_of_measurement) 
+                                                                                                  :  $scope.states[entityId].attributes.friendly_name
```
rchl: Instead of repeating `$scope.states[entityId]` in multiple places you could create variable for it. But actually, it's better to take that information from the response data instead as, as far as I can see, it's there too.

README.md
```diff
@@ -383,6 +383,16 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
     * (optional)
     */
    customStyles: Object || Function
+
+   /* Object containing history settings */
+   history: { // If this is present in a tile, a history popup is created on long-press
```
rchl: Please call it "secondary action" instead of long-press

scripts/controllers/main.js
rchl: Well, each data point includes the data but it might feel a bit weird to get that data from the first data point. So maybe just create variable for `$scope.states[entityId]`.

akloeckner: I used the data, because one might actually use `history.entity`.

scripts/models/api.js
akloeckner: I fixed with the actual error codes from the API.

Successful calls will return status code 200 or 201. Other status codes that can return are:
-    400 (Bad Request)
-    401 (Unauthorized)
-    404 (Not Found)
-    405 (Method not allowed)

https://developers.home-assistant.io/docs/en/external_api_rest.html

scripts/models/api.js
akloeckner: They are actually optional. So, I'd leave this as is, because it can be used for other things later on then...
- no `startDate` = one day
- no `filterEntityId` = all entities. But yes, only one is used in the chart.

scripts/models/api.js
akloeckner: I added padding. The closing button is the same for all popups. So, making it smaller is out of scope for me, right now.

rchl:
Could you give me access to push to branch? I want to do some minor change but avoid long and annoying back and forth.

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

akloeckner:
That should work already, if I interpret the checkmark correctly...

rchl:
Hmm, I guess being a collaborator, I might not be allowed to push...

rchl:
Can't make it work. I will submit improvements in a separate PR.

Thanks for your work on this and patience with a picky reviewer. :)

akloeckner:
> Thanks for your work on this and patience with a picky reviewer. :)

You're welcome!

PR #275, opened 2019-10-20
Fix item's state behavior

akloeckner:
This alters the `item.state(s)` fallback logic, in order for the following to work:

```
state: function (item, entity) {
   return entity.attributes.can_cancel ? entity.state : null;
},
```

The purpose is to remove the state field from tiles of one-shot scripts, which have no meaningful state.

The new fallback falls in line with the 'highly customizable' objective of TileBoard:
* if the user provides `state` or `states`, the field is used - no matter what, so also `null`, `false`, `{}` or similar results.
* if the user provides sth, which is not a function, string or object, it is simply passed on as the state.
* if none of the fields are provided, the `entity.state` is returned.

I believe this is very similar to the old logic, but more consistent and more customizable. However, people might have relied on the old behavior; e.g. (whyever) providing `state` and `states` together using the latter as a fallback for empty result of the former.

For you to decide, if this change is useful. I believe it is - for the sake of consistency and customizablility. Hence the PR. ;-)

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -451,32 +451,31 @@ function MainController ($scope, $location) {
    $scope.entityState = function (item, entity) {
       if(item.state === false) return null;
 
-      var res;
-
-      if(item.state) {
+      if(typeof item.state != 'undefined') {
          if(typeof item.state === "string") {
             return parseString(item.state, entity);
          }
          else if(typeof item.state === "function") {
-            res = callFunction(item.state, [item, entity]);
-            if(res) return res;
+            return callFunction(item.state, [item, entity]);
+         }
+         else {
+            return item.state;
          }
       }
 
-      if(item.states) {
+      if(typeof item.states != 'undefined') {
          if(typeof item.states === "function") {
-            res = callFunction(item.states, [item, entity]);
+            return callFunction(item.states, [item, entity]);
          }
          else if(typeof item.states === "object") {
-            res = item.states[entity.state] || entity.state;
+            return item.states[entity.state] || entity.state;
          }
-
-         if(res) return res;
+         else {
+            return item.states;
+         }
       }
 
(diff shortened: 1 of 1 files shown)


scripts/controllers/main.js
```diff
@@ -451,32 +451,31 @@ function MainController ($scope, $location) {
    $scope.entityState = function (item, entity) {
       if(item.state === false) return null;
 
-      var res;
-
-      if(item.state) {
+      if(typeof item.state != 'undefined') {
```
rchl: Should use  `!==` and can also use shorter `item.state !== undefined`.

scripts/controllers/main.js
```diff
@@ -451,32 +451,31 @@ function MainController ($scope, $location) {
...
          if(typeof item.states === "function") {
-            res = callFunction(item.states, [item, entity]);
+            return callFunction(item.states, [item, entity]);
          }
          else if(typeof item.states === "object") {
-            res = item.states[entity.state] || entity.state;
+            return item.states[entity.state] || entity.state;
+         }
+         else {
+            return item.states;
```
rchl: I have doubt about having this branch. `states`, according to documenation, `Map a state from HomeAssistant to a different value for TileBoard to display.` So returning just plain value that it's defined to kinda doesn't make sense, I think.

Also, according to README, only object and function are supported so `string` or whatever else shouldn't be supported. Maybe instead trigger a `console.warn` here and fallback to entity state...?

akloeckner: Agreed. It would additionally then simply duplicate `item.state` for that case.

scripts/controllers/main.js
akloeckner: Fixed strict comparison. 

I kept `typeof` though, as it handles `ReferenceError` well and can't be broken by defining a `var undefined`.

PR #277, opened 2019-10-21
More robust default location

akloeckner:
Assuming HA is on the same machine as TileBoard. Also for special cases of SSH tunneling or the like. Also, TileBoard should then work just by copying the example to `config.js`. Except for the case of an `https` server, obviously.

Also converted DOS line endings to NIX in a separate commit to allow cherry-picking. Assuming most people are on some kind of NIX system. This bugged me in the very beginning and I had to search for some time how to convert it.


config.example.js
```diff
@@ -13,8 +13,8 @@ var CONFIG = {
    entitySize: ENTITY_SIZES.NORMAL, //SMALL, BIG are available
    tileSize: 150,
    tileMargin: 6,
-   serverUrl: "http://localhost:8123",
-   wsUrl: "ws://localhost:8123/api/websocket",
+   serverUrl: "http://"+window.location.hostname+":8123",
```
rchl: Should we also use `location.protocol`?

(BTW. I think it's fine to skip `window.` part.)

resoai: If we are to use `location.protocol`, we could also do `ws://` vs `wss://`. Please also make sure there are spaces before and after `+`.

akloeckner: If we were to use the protocol, it would look like this:
```
   serverUrl: location.protocol +                       "//" + location.hostname + ":8123",
   wsUrl:     location.protocol.replace('http', 'ws') + "//" + location.hostname + ":8123/api/websocket",
```

A bit messy already. And I believe, if people know how to set up HTTPS plus the necessary browser exceptions, they should also feel comfortable to add the extra `s`'s. 

Plus: they are likely to also change the `location.port`. Which is even more messy to catch the empty results in case of a standard port.

What do you think?

rchl: I won't argue about that more as there is no "right way" here as it's basically guessing and trying to cater for the most common case.

Please just fix style issues and I'll be OK with it.

akloeckner: Check!

rchl:
Merged manually with slight history rewriting and style change. Thanks!

PR #280, opened 2019-10-25
Don't clear access token on intermittent connection problem

rchl:
When HomeAssistant is restarting or just there is a problem with
connection, TileBoard assumed that user needs to get authenticated
and tried to fetch new token after deleting current one.

Don't do that unless the response code is in 4xx range.
Per HomeAssistant documentation[1], it will return 400, 401 or 403
when token is expired or there is other authentication error.

Also don't react to state changes until response is in DONE state.

[1] https://developers.home-assistant.io/docs/en/auth_api.html#making-authenticated-requests


resoai:
Yet again, thank you!

Issue #229, opened 2019-05-18
[Feature Request] long press


JesseWebDotCom:
I'd love to have the ability to prevent clicks and only have a long press.  This is particularly important for cover tiles, where I don't want someone accidentally opening the garage door because they were swiping/scrolling through the dashboard.

To start, perhaps you can quickly add this to the custom tile (ex. action = click press, action_long = long press)?

resoai:
This breaks functionality with light tile which has separate functionality for long press and is not an expected behaviour. You are free to override the default handlers in main.js for your own setup though.

JesseWebDotCom:
Is there any way to override handlers without modifying core TileBoard files?  I'm concerned about hacking up core files, updating when you have a new release, and I lose my changes.

Thanks

resoai:
There will hardly be any changes to that file in future.

JesseWebDotCom:
I was able to add long press to TYPES.CUSTOM with the code below.  Perhaps you can incorporate this into the official release for others:

MAIN.JS
```
   $scope.entityLongClick = function ($event, page, item, entity) {
      $event.preventDefault();
      $event.stopPropagation();

      switch (item.type) {
         case TYPES.LIGHT: return $scope.openLightSliders(item, entity);
         // ONLY ADDED THE LINE BELOW
         case TYPES.CUSTOM: return $scope.customTileActionLong(item, entity);
      }

      return false;
   };

   // ADDED THIS FUNCTION
   $scope.customTileActionLong = function (item, entity) {
      if(item.action_long && typeof item.action_long === "function") {
         callFunction(item.action_long, [item, entity]);
      }
   };   
```

CONFIG.JS (Example on how to make an open/close garage door tile REQUIRE a long press)
```
{
   type: TYPES.CUSTOM,
   state: false,
   id: 'cover.garage',
   title: 'Garage',
   icons: {
      open: 'mdi-garage-open',
      opening: 'mdi-garage-alert',
      closing: 'mdi-garage-alert',
      closed: 'mdi-garage'
   },
   action_long: function (item, entity) {
      var service_name = 'close_cover';
      if (entity.state === 'closed') {
         service_name = 'open_cover';
      }

      Api.send({
         type: "call_service",
         domain: "cover",
         service: service_name,
         service_data: {
            entity_id: item.id
          }
      });
   }, 
```

JesseWebDotCom:
[...]

PR #282, opened 2019-10-26
Allow override of `item.action` and `item.longClickAction`

akloeckner:
This caters for ideas as described in #274. If we can always override the `item`'s `action` or `longPressAction` we can make more fancy things. And TileBoard becomes actually more "customizable".

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,18 @@ Tile Object. [Click here for some real life examples](TILE_EXAMPLES.md)
    */
   icon: 'mdi-phone'
 
+  /* action: Define a custom action on click
+   * You can override the default action for any tile type.
+   * This function will be evaluated, when the user clicks the tile.
+   */
+  action: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
+
+  /* longPressAction: Define a custom action on long press
+   * You can override the default action for any tile type.
+   * This function will be evaluated, when the user long-presses the tile.
+   */
+  longPressAction: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
+
   /* bg: Link to a background image for the tile
    * @ and & prefixes are explained below
    */
--- a/README.md
+++ b/README.md
@@ -286,11 +286,11 @@ Tile Object. [Click here for some real life examples](TILE_EXAMPLES.md)
    */
   action: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
 
-  /* longPressAction: Define a custom action on long press
-   * You can override the default action for any tile type.
+  /* secondaryAction: Define a custom secondary action (on long press)
+   * You can override the default secondary action for any tile type.
    * This function will be evaluated, when the user long-presses the tile.
    */
-  longPressAction: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
+  secondaryAction: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
 
   /* bg: Link to a background image for the tile
    * @ and & prefixes are explained below
--- a/TILE_EXAMPLES.md
+++ b/TILE_EXAMPLES.md
@@ -80,7 +80,7 @@ Manually trigger an automation
 ```
 
 #### CUSTOM
-The custom tile type allows you to fire javascript commands on click/tap.
+The custom tile type does not have handling for any specific entity types. It can be used to, for example, trigger custom actions on pressing.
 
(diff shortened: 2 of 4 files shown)


akloeckner:
PS: We can argue about the option's name. I was struggling between `longClickAction` and `longPressAction`. The former was used in TileBoard before. The latter seems to be more "common language". Another option would be to use `pressAction`.

rchl:
Can you add documentation, making it clear that this applies to all tile types, not only custom.

akloeckner:
Agreed and fixed.

I also fixed the name of `entityLongClick` for consistency.

resoai:
Perhaps entityClick needs to become entityPress?

akloeckner:
I don't know. I'm quite sure, I wouldn't say “I long-clicked a tile“. But for click, it's also `onClick` for example.Where it was `hmPress` for long press. So, maybe “press“ already implies “long“?

But honestly, I don't know what is the correct wording in the app developer world. ;-)

resoai:
I guess it is “click” with mouse and “tap” with a finger.

akloeckner:
For the sake of consistency, shall we just use the words already established by the HTML attributes? 

https://github.com/resoai/TileBoard/blob/6f1e8d76139c7efaca23f66ed61f26faefae4a61/index.html#L455

That would be “click“ for short and “press“ for long. That's how it is in this PR now. For the user, there's no ambiguity, because `action` is clear without any of these words. And `longPressAction` includes “long“ for clarity.

But, as ever, it's your project, so, I'll change it to whatever you consider best.

akloeckner:
> Can you add documentation

Could have a quick look to check if this is really the shortest way to use UI functions in the custom actions? It seems a bit too complicated for my taste...

```
action: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
```

resoai:
> > Can you add documentation
> 
> Could you have a quick look to check if this is really the shortest way to use UI functions in the custom actions? It seems a bit too complicated for my taste...
> 
> ```
> action: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
> ```

To the best of my knowledge it is the only way. We never through about necessity to call those functions from here hence the awkwardness. Her his what I did for `openPage`:

https://github.com/resoai/TileBoard/blob/master/scripts/controllers/main.js#L2039

akloeckner:
Ah, I see. But we shouldn't provide such shortcuts for all possible functions. Whoever wants to use whatever function from `$scope` will have to implement their own shortcut then. Thanks for checking! 

Do you want to change some click/tap/press names still, or shall we keep it as is for the time being?

resoai:
I really don’t mind what we have at the moment. Let’s see what our guru @rchl says :)

akloeckner:
:-) thanks.

akloeckner:
As a note and for automated closing, this fixes #229, by allowing to specify an empty
```
action: function(item, entity) {}
```

README.md
```diff
@@ -280,6 +280,18 @@ Tile Object. [Click here for some real life examples](TILE_EXAMPLES.md)
...
+   * You can override the default action for any tile type.
+   * This function will be evaluated, when the user clicks the tile.
+   */
+  action: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
+
+  /* longPressAction: Define a custom action on long press
+   * You can override the default action for any tile type.
+   * This function will be evaluated, when the user long-presses the tile.
+   */
+  longPressAction: function(item, entity) {return this.$scope.openPopupIframe(item, entity);}
```
rchl: How about `secondaryAction` then, for parity?
That would leave-out semantics of how they are triggered.
Could even have them triggered in different ways on different platforms (long press on desktop is not that common for example). But that would be out of scope of this change of course.

rchl:
The documentation for CUSTOM tile now says:
> The custom tile type allows you to fire javascript commands on click/tap.
which is weird since all tiles can do that now. It could say, for example:

`The custom tile type doesn't have handling for any specific entity types. It can be used to, for example, trigger custom actions on pressing.`

And the code could also include an example for using `secondaryAction` (or whatever we decide to call it).

README.md
resoai: secondaryAction it is than!

rchl:
Merged manually as c28106f4925613d3ccaba990ded5806c7a3a9f18 with some minor corrections - updated commit message and move documentation in README a bit lower.

akloeckner:
You're welcome as ever. Happy to contribute.

PR #284, opened 2019-11-10
Convert Api class into angular.js service (provider)

rchl:
This allows for injecting angular objects like `$http` which:
a) other changes will depend on.
b) currently used plain XHR should be replaced with it.

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,13 +47,13 @@
          'scripts/vendors/color-picker.js',
          'scripts/vendors/angularjs-gauge.js',
 
-         'scripts/models/api.js',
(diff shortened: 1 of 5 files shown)


rchl:
Enable `Hide whitespace changes` in github's diff view to make diff easier to review.

rchl:
@resoai BTW, related to future changes, any particular reason code doesn't use promises but instead uses callbacks? I belive angular.js polyfills promises if they are not available.

resoai:
I think that it was simply an old habit :) I don’t mind promises.

PR #289, opened 2019-11-11
Fix usage of return value from $q

### LINKED ISSUE #288: Issue with authentivation token manipulation in last commit b228791
Hello,

it seems that the latest version ( b228791 ) of Tileboard has an issue with the auth token manipulation with the following message : 
```TypeError: Cannot read property 'then' of undefined```

The error seems to come from the models/api.js fil here when calling **this.getToken().then(....)**.

I am far from a JS expert, but i may try to have a look into it soon.
Currently I took an older commit and it works great.

Grm

rchl:
Resolves #288

### DIFF
--- a/scripts/models/api.js
+++ b/scripts/models/api.js
@@ -337,10 +337,8 @@ App.provider('Api', function () {
       };
 
       $Api.prototype._getToken = function () {
-         var deferred = $q.defer();
-
          if(this._configToken) {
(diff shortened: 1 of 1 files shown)


resoai:
Please feel free to merge at will. I trust you :) I’m only able to review stuff visually on my phone at the moment, so it’s is not worth it anyway.

PR #293, opened 2019-11-16
Chart improvements

rchl:
@akloeckner


scripts/init.js
```diff
@@ -1,3 +1,10 @@
+if(!window.CONFIG) {
+   var error = 'Please make sure you have "config.js" file and it\'s a valid javascript!\n' +
+      'If you running TileBoard for the first time, please rename "config.example.js" to "config.js"';
+
+   alert(error);
```
resoai: I’ve just realised that it should be “If your are running...”

rchl: Will fix. Note that I'm planning a couple of more changes so please wait with merging. :)

resoai: Please feel free to merge :) As I mentioned before, I completely trust you :)

rchl:
This doesn't look intuitive to me:
off-state
This is an entity that was only in 'off' state for the whole period of shown history. And it looks like it was on  to me, if you don't read into labels.

akloeckner:
> This doesn't look intuitive to me:

It could be completed in the yAxis' label computation for things like on/off, home/not_home etc.

rchl:
> > This doesn't look intuitive to me:
> 
> It could be completed in the yAxis' label computation for things like on/off, home/not_home etc.

Can you think of a specific solution?
I wouldn't want to hardcode possible values because, in theory, the value could be any string.
It could be that a line graph is just not a good fit for state values like those. But Chat.js doesn't seem to have a more appropriate chart for this case, I think.

akloeckner:
We could hardcore just a few common values. Suppose, e.g., if any in `[on, off`] is in the state history, then extend labels by the rest in `[on, off]`. Same for other sets of states, we know of. But it will probably stay incomplete forever.

Maybe there is at least some comprehensive list of binary states. Such as here: https://www.home-assistant.io/integrations/binary_sensor/. Although I think, that specific list is (1) a bit overkill and (2) not complete, I'm missing home e.g...

akloeckner:
> It could be that a line graph is just not a good fit for state values like those. 

I'm also not very happy with the graphs for that in HA itself. Short changes can only be seen in the tooltip, if at all...

rchl:
> We could hardcore just a few common values. Suppose, e.g., if any in `[on, off`] is in the state history, then extend labels by the rest in `[on, off]`. Same for other sets of states, we know of. But it will probably stay incomplete forever.

Handling most common states like that would possibly be worth it although I'm usually not a fan of such half-baked solutions. States can be anything really, for example for media devices it can be "on, off, playing, paused, unavailable". For Sun state it can be "Below/Above horizon", and the list goes on...

> > It could be that a line graph is just not a good fit for state values like those.
> 
> I'm also not very happy with the graphs for that in HA itself. Short changes can only be seen in the tooltip, if at all...

That's true. It's not ideal for things like door sensors where it can be "on" only for a couple of seconds a day...

I won't do anyting about that issue then. Current solution works, even if confusing sometimes.

akloeckner:
In case you missed my edit: Maybe a good and simple solution for your specific problem would be to have the single y label be centred vertically in the graph? Not sure how to do it though.

rchl:
> In case you missed my edit: Maybe a good and simple solution for your specific problem would be to have the single y label be centred vertically in the graph? Not sure how to do it though.

Could do that by adding an empty label both before and after existing label, if only one present.
But I think I will actually add handling for the on/off case at least. It's the most important one really.

akloeckner:
Hmm. I'm not familiar with code comments as it seems. I put one on your last commit.

akloeckner:
I just opened #294. If you can think of a generic solution to add those directives, maybe you can incorporate the fix here...

rchl:
> And why not add an else branch to include the bogus two empty labels before and after? You'd have to move the `sort` operation up then, too.

Did as you suggested.

rchl:
BTW. There seems to be no need to provide `yAxes.labels` when type is `linear`. Chart.js figures out the scale automatically from available data. Do you remember if you saw some case when it's needed or you just did it because it seemed to make sense @akloeckner ?

akloeckner:
I believe, I did it, because I didn't know how to conditionally add the field to the struct only for categorial data.

resoai:
What a great effort of both of you guys! Well done!

rchl:
I'm ready with the changes. Feel free to look through and point out if anything looks wrong.

One thing that I'm not certain about is changing the position of the close button for all popups. I haven't tested all those popup cases apart from the history popup itself.

rchl:
I was also looking into supporting multiple entities in one chart, but that is proving a bit difficult, even with same-unit entities.

akloeckner:
> I was also looking into supporting multiple entities in one chart, but that is proving a bit difficult, even with same-unit entities.

Yes... The x labels are the problem. The `datasets` can be overridden. But that is some bigger effort, I believe.

> Feel free to look through and point out if anything looks wrong.

I'll check again in the coming days. But I think I'm up to date and I haven't seen anything disturbing so far.

akloeckner:
:+1:

PR #295, opened 2019-11-18
Add title bar area to popups

rchl:
Suggestion on how to make the close button look better in popups - introduced a titlebar area that fits both the close button and entity's title. Tell me what you think.
Changes in main.less might be a bit messy as I don't fully understand currently used convention (is it BEM, custom convention or a mix?).
@akloeckner

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,6 +77,13 @@
         ng-if="activeCamera && (entity = getItemEntity(activeCamera))">
 
       <div class="camera-popup-container">
+         <div class="camera-popup--title">
+            <div class="camera-popup--close" ng-click="closeCamera()">
+               <i class="mdi mdi-close"></i>
+            </div>
+            {{ entityTitle(activeCamera, entity) }}
+         </div>
+
          <div class="camera-popup--list">
             <div class="camera-popup--list-item"
                  ng-click="openCamera(item)"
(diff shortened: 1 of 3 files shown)


rchl:
Example with camera and history popups:
camera
sensor
(camera popup looks horrible on phone display BTW, but it's not a new issue)

resoai:
It was supposed to be BEM. Should the title perhaps be overridable?

rchl:
That's better indeed.
Screenshot 2019-11-19 at 08 11 08

akloeckner:
It looks very nice. Maybe this can make the chart legend obsolete? I’d support a user-changeable title, but it seems that is already the case? (The title being different than the legend.)

If more such common pop up things will be introduced: would it make sense to refactor the code to have the title bar be actually common for all popups?

resoai:
> It looks very nice. Maybe this can make the chart legend obsolete? I’d support a user-changeable title, but it seems that is already the case? (The title being different than the legend.)
> 
> If more such common pop up things will be introduced: would it make sense to refactor the code to have the title bar be actually common for all popups?

I think it should be common to all popups.

rchl:
> I’d support a user-changeable title, but it seems that is already the case? (The title being different than the legend.)

It uses `entityTitle` function which takes tile's `title` or `friendly_name` from entity. So yes, one can override it but it will be the same title that is shown on the tile itself then. To me, that seems fine. We could, of course, have another level of override for the popup title specifically but maybe let's not try that for now? We can add later if really needed.

> > If more such common pop up things will be introduced: would it make sense to refactor the code to have the title bar be actually common for all popups?
> 
> I think it should be common to all popups. I would drop the legend since we only have one entity and title on top.

About dropping the legend - legend's text has actually more information (full entity name and dataset unit, if there is one). The title of the popup is either `title` from the `tile` or entity name. So it wouldn't be the same. Also, I think I'm gonna look next into supporting multiple entities so we would need the legend anyway.

As for unifying popups, I think it should be a separate PR. And those would possibly be breaking changes for some  as we would have to change class names.

resoai:
Agreed! Shall I merge it than?

rchl:
> Agreed! Shall I merge it than?

Give me some time, please. I want to get familiar with BEM and see what I've messed up in css. :)

akloeckner:
> Agreed

Me too. ;-)

rchl:
> It was supposed to be BEM.

I think it's not conforming to BEM very well right now.

In BEM the convention is (according to https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/):
```
.site-search {} /* Block */
.site-search__field {} /* Element */
.site-search--full {} /* Modifier */
```
And there is CSS like this right now:
```
.history-popup {
   // ...
   &-container {
```
which compiles to:
```
.history-popup { ... }
.history-popup-container { ... }
```
So that would mean that both are blocks (components) but I'd say `history-popup-container` should be an `element` thus it should be called `history-popup__container`. It's a child of `history-popup` and I wouldn't call it a separate component.

rchl:
I've learned enough about BEM now and updated my changes accordingly, to follow it better, but the existing code doesn't really follow BEM very well so I've also tried to be consistent with existing code.
It's done from my side.

resoai:
@rchl @akloeckner Thanks so much for all your effort!

PR #297, opened 2019-11-19
Fix/chart close button and overflow

rchl:
I suggested checking per-commit as those are two separate changes and also have a lot at the commit message where I explain my intentions.
@akloeckner

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -1043,21 +1043,21 @@ camera-thumbnail div {
   line-height: 50px;
   overflow: hidden;
   text-overflow: ellipsis;
-  padding-left: 10px;
+  padding-right: 10px;
   white-space: nowrap;
 }
 .camera-popup-close,
 .door-entry-popup-close,
 .iframe-popup-close,
 .history-popup-close {
-  float: right;
+  float: left;
   text-align: center;
   color: rgba(255, 255, 255, 0.6);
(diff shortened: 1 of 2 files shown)


akloeckner:
To the close button:

Having close at the left is iPhone habits, right? Android and Windows would be right side? Do you have any preferred way in this project? I tend to like the idea of a left side close button better.

To the chart height:

I did quite some trial and error to make that work. And one goal was to fix overflow using `max-height ` in landscape mode. So, this should work already. It did for me at least. 

However, it seems much cleaner to use `@media ` explicitly. Except for the `calc`s. Are they necessary? 

As I'm traveling and can't check: what does that do to non-mobile displays? I'd suggest to have those also at 100% then, since it's also landscape? Which I'd like, because it wastes less space.

rchl:
Close button: It's typical to have the close button on the left on iPhones and Macs (IOS and OSX) yes. But HomeAssistant also has it on the left, regardless of the platform.

Chart height: If you are saying that it should have worked before then maybe I broke it with the addition of title bar. It complicates things a bit since it takes 50px of the height (that's why we need the calc also, to subtract that 50px).
With non-mobile displays it will be at 100% since those are in landscape. Basically it switches to 100% when width becomes bigger than height.

resoai:
Thanks you!

PR #299, opened 2019-11-20
Support multiple entities in charts

rchl:
This adds support for showing multiple entities in the chart.
- Nothing should change for single entities
- With multiple entities, all using the same value unit (like degrees or watts), there will be a single y-axis scaled to accommodate all values. That's a perfect case.
 - With multiple entities with different value units or with non-continuous data, there will be multiple y-axes added. That's not an ideal situation and not really possible to handle perfectly. Ideally, there should be multiple graphs created for those cases. But it's sort of usable and if users want to do that, it's up to them, I'd say.
@akloeckner

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,11 +382,11 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
     * as a function function (item, entity) { return { 'background-color': '#FF0000' } }
     * (optional)
     */
-   customStyles: Object || Function
+   customStyles: Object || Function,
 
    /* Object containing history settings */
    history: { // If this is present in a tile, a history popup is created on secondary action
-      entity: 'sensor.temperatur_innen_gefiltert', // Entity ID to render history for. Default: entity id of the tile itself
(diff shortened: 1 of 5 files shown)


rchl:
Two temperature sensors:
double-temp

Light switch + watt power usage:
light_and_watt

Mayhem:
mayhem

resoai:
Thank you! I suppose there is no option in Chart.js of adding a little padding after legend and before the graph?

rchl:
> Thank you! I suppose there is no option in Chart.js of adding a little padding after legend and before the graph?

I was looking for it before but no, didn't find anything.
I suppose we could disable canvas legend and draw our own (like HA does).

resoai:
According to StackOverflow:

`
Chart.Legend.prototype.afterFit = function() {
    this.height = this.height + 50;
};
`

rchl:
Btw. Things don't look good on a small screen like a phone and multiple entities. Legend takes too much space (even without changing padding)...

resoai:
> Btw. Things don't look good on a small screen like a phone and multiple entities. Legend takes too much space (even without changing padding)...

We could get rid of them and make our own as you suggested. What would you prefer to do?

PR #304, opened 2019-11-26
Fix date formatting in the graph

rchl:
Since date was left in the string form (as it comes from the API), the
local date formatting was not applied in the graph (in a tooltip).

@akloeckner


rchl:
Before:
Screenshot_20191126-153954_Chrome

After:
Screenshot_20191126-152943_Chrome

resoai:
Thank you!

PR #314, opened 2019-12-02
Warn for unknown entity only once

### LINKED ISSUE #311: entity not found
If an entity is not found, the NOTY is not dismissable. It just comes back. I believe, the error should only be thrown once.

For me this is an issue, because my chromecast is not always running. So after a restart of HA, it is missing until I start the TV for the first time.

akloeckner:
fixes #311

### DIFF
--- a/scripts/models/noty.js
+++ b/scripts/models/noty.js
@@ -147,9 +147,9 @@ var Noty = (function () {
    };
 
    Noty.getById = function (id) {
-      for(var i = 0; i < Noty.noties.length; i++) {
-         if(Noty.noties[i].id === id) {
-            return Noty.noties[i];
+      for(var i = 0; i < Noty.notiesHistory.length; i++) {
+         if(Noty.notiesHistory[i].id === id) {
+            return Noty.notiesHistory[i];
          }
       }
 
--- a/scripts/models/noty.js
+++ b/scripts/models/noty.js
@@ -147,13 +147,23 @@ var Noty = (function () {
    };
 
    Noty.getById = function (id) {
-      for(var i = 0; i < Noty.notiesHistory.length; i++) {
-         if(Noty.notiesHistory[i].id === id) {
-            return Noty.notiesHistory[i];
+      for(var i = 0; i < Noty.noties.length; i++) {
+         if(Noty.noties[i].id === id) {
+            return Noty.noties[i];
          }
       }
 
       return null;
+   };
+
+   Noty.getHistoryById = function (id) {
+      var history = [];
+      for(var i = 0; i < Noty.notiesHistory.length; i++) {
+         if(Noty.notiesHistory[i].id === id) {
+            history.push(Noty.notiesHistory[i]);
+         }
+      }
+      return history;
    };
 
    Noty.remove = function (noty) {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -2145,7 +2145,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    }
 
    function addError (error) {
-       if(!CONFIG.ignoreErrors) Noty.addObject({
+      if(!CONFIG.ignoreErrors) Noty.addObject({
          type: Noty.ERROR,
          title: 'Error',
          message: error,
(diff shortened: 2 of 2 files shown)


scripts/models/noty.js
```diff
@@ -147,9 +147,9 @@ var Noty = (function () {
    };
 
    Noty.getById = function (id) {
-      for(var i = 0; i < Noty.noties.length; i++) {
-         if(Noty.noties[i].id === id) {
-            return Noty.noties[i];
+      for(var i = 0; i < Noty.notiesHistory.length; i++) {
```
rchl: You're changing the behavior of `Noty.addObject` with this change since it also uses `getById`. You could instead add a new function like `Noty.addObjectIfNew` that would have the logic of checking through history.

akloeckner: Yes. I'm aware of this. But I couldn't think of any side effects. Can you? If not, I'd prefer the simple change. Which even can be useful for other things, possibly.

rchl:
The side effect of that you apply that behavior to all notifications, not only ones about entity not found. We don't want those to be only triggerable once.

scripts/models/noty.js
rchl: The side effect of that is that you're applying that behavior to all notifications, not only ones about entity not found. We don't want those to be only triggerable once.

akloeckner: The actual "apply only once" magic is not here. It's in `main.js`.

The side effect of this (supporting) change here is mainly, that Noties (which have already disappeared) will be copied (by reference) when rethrown (instead of recreated). This should not change the actual Noty shown. It will, however, change the Noty in the `notiesHistory`, which is not used anywhere anyways.

This is why I figured there is no side effect. But I accept being reasoned wrong. ;-)

rchl: > Noty.add(Noty.ERROR, 'Test error', 'Message body', null, null, 'custom-id')

No, `Noty` code also follows that logic with your changes.
Try triggering this twice from a console (closing notification after first trigger):
```
Noty.add(Noty.ERROR, 'Test error', 'Message body', null, null, 'custom-id')
```
With your changes you can only trigger it once in session.

akloeckner: Ups, you're right... a pushed a fix. With a less extensive solution than proposed, however...

scripts/models/noty.js
```diff
@@ -156,6 +156,16 @@ var Noty = (function () {
       return null;
    };
 
+   Noty.getHistoryById = function (id) {
```
rchl: I think the name is not ideal but then maybe let's think if it really needs to return an array of notes.
Do you see a use case for looking at historical notes? I don't see one right now at least. Instead, I would suggest calling it `hasSeenNoteId` and just return `true` if note id is found. That would be enough for your use case.

rchl:
On a more general note, that change will hide information that maybe users would like to see. It's fairly opinionated change based on your use-case alone. So I think I'm not 100% convinced about it...

What makes it less impactful of a change is the fact that on reload of TileBoard, the history will be wiped out, so the effect of this change will not persist for long for most people (on phones the browser will be removed from memory frequently and the page will reload on re-opening). So I'm not worried too much that this change. But that fact will also make the change less effective for you, I can imagine.

Maybe the ultimate solution would be to have a special state for tiles that are missing entity instead of triggering a notification. I think that would be a better UX.

All that said, I think I'm fine with merging this change now and potentially improving UX later.

resoai:
How about simply adding `ignoreErrors` for a tile?

rchl:
> How about simply adding `ignoreErrors` for a tile?

Personally I'm not a fan of adding more options.
Also, I also was annoyed by the persistence of that notification before but I wouldn't want to have any indication whatsoever that something went wrong. So part of myself is supporting this change. :)

resoai:
Agreed. If there are no more changes required, than we could merge.

scripts/models/noty.js
akloeckner: Debugging could be a use case. But it's marginal. I just tend to prepare for more general use cases. Your suggestion sounds just as good to me.

akloeckner:
> additional option
I agree. That would make thing complicated.

> users might want to see this
I doubt that. The noty does never disappear, because apparently the view is constantly updated. That means the noty permanently hides a third of the screen. Who would want to see that? But probably you're right. There always is someone with a particular use case.

> the page reloads
Yes, that makes it less effective. But I'm fine to click once I load the page. I'm not with the persistent noty.

> Better UX with special tile state
Yes! That would be the best...

> Summary...
I agree. Let's keep it as is: a bit hacky fix, even with returning the array. And let's keep the special tile state in mind.

rchl:
I've  integrated this manually with my suggested change.

PR #317, opened 2019-12-04
Toggle pages with panning instead of swiping

rchl:
Panning provides much better and fluid experience and is easier to discover.

I'm quite happy with this change - been using it for some time myself.

There is one thing worth noting about this implementation and related side effect.

On panning, page position is updated (through CSS transform) manually to follow finger motion.
On lifting the finger, I'm finishing the movement using CSS transition.
While that makes implementation much simpler, there is a slightly noticeable "hickup" when that transition happens. It looks a bit like a frame drop. That part could be improved but it would require some fancy math formula to calculate the acc/deceleration of transition instead of relying on CSS transitions. So I'm not sure how worth it is.

Maybe try yourself and check if it's bothersome (now that I mentioned it maybe it will be more than otherwise would be. ;)).

### DIFF
--- a/index.html
+++ b/index.html
@@ -70,7 +70,7 @@
 <body ng-controller="Main" ng-class="getBodyClass()">
 
 <div ng-if="ready" class="page-container"
-     hm-swipe="onPageSwipe($event)"
+     hm-pan="onPagePan($event)"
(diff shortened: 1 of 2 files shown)


resoai:
It is actually not that bad. Perhaps one day (with a new project structure) this can be improved yet it is perfectly find for now. Thanks!

PR #319, opened 2019-12-05
Trigger popup history on long press if no other handlers

rchl:
Trigger history popup on long press if entity doesn't have any other handler.

I think it makes sense but you decide.
I like to be able to see history for tiles like "gauge" or even "switch" to be able to see historical/usage data.

@akloeckner

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -84,7 +84,11 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
       switch (item.type) {
          case TYPES.LIGHT: return $scope.openLightSliders(item, entity);
-         default: return $scope.openPopupHistory(item, entity);
+         default: {
+            if (entity && entity.entity_id) {
+               return $scope.openPopupHistory(item, entity);
+            }
+         }
       }
    };
 
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -84,7 +84,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
       switch (item.type) {
          case TYPES.LIGHT: return $scope.openLightSliders(item, entity);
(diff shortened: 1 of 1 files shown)


resoai:
I'm just unsure about it being `default` since we have camera, door entry, alarm (perhaps something else) which open their own popup and not on the long press.

rchl:
> I'm just unsure about it being `default` since we have camera, door entry, alarm (perhaps something else) which open their own popup and not on the long press.

 - I can imagine history could be useful for alarm tile since it should show a history of states then.
 - I can change it so that it won't trigger if `entity` is not defined which should cover DOOR_ENTRY (nothing should happen)
 - as for the camera, I'm not sure if those have different states. Maybe they have things like motion/recording/idle or something and then it would also be useful?

resoai:
> 
> 
> > I'm just unsure about it being `default` since we have camera, door entry, alarm (perhaps something else) which open their own popup and not on the long press.
> 
>     * I can imagine history could be useful for alarm tile since it should show a history of states then.
> 
>     * I can change it so that it won't trigger if `entity` is not defined which should cover DOOR_ENTRY (nothing should happen)
> 
>     * as for the camera, I'm not sure if those have different states. Maybe they have things like motion/recording/idle or something and then it would also be useful?

I agree about alarm and door entry (this would also cover iframe and iframe popup). As for the camera, mine doesn't expose anything but `idle` and `unavailable`, the later could perhaps be useful for some people, so let's keep it!

resoai:
Cool! Thanks!

akloeckner:
Also, maybe the condition should be in `openHistoryPopup`. Where my other condition might already reside...

rchl:
@akloeckner Such condition wouldn't make sense in context of what this change does.

You can disabled popup by overriding `secondaryPress` function. We could also add something like `history: { disable: true }` but it might be too much.

PR #322, opened 2019-12-08
activate panning only, if not scrolling

akloeckner:
otherwise panning would be started partly when scrolling through a page, which looked awkward


akloeckner:
Code picked from here: https://github.com/hammerjs/hammer.js/issues/771

rchl:
It's some old Safari right? I can't see a problem in iOS 13.

akloeckner:
It's my firefox mobile on android. Which should be current. 

Just to be sure, I tried on my old iPad. The change does not fix the awkward behavior on it...

akloeckner:
I retried on my firefox. The fix doesn't seemt to be a 100% fix there either. I'll close this PR and open an issue instead.

PR #334, opened 2020-01-19
New "image" tile

christophebaraer:


scripts/controllers/main.js
```diff
@@ -942,6 +942,15 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
       return null;
    };
+    
+    
+   $scope.itemCustomURL = function (item, entity){
+       if (typeof item.imageURL === 'function') {
+         return callFunction(item.imageURL, [item, entity]);
+      }
+
+      return item.imageURL;
```
rchl: Since it's an IMAGE tile, it should be clear from the context that it's the *image* URL so maybe just call the property `url`?

TILE_EXAMPLES.md
```diff
@@ -633,3 +633,15 @@ This is a custom tile which can be used for displaying values from different sen
    },
 }
 ```
+
+
+#### IMAGE
+![IMAGE](images/tile-screenshots/IMAGE.png)<br>
```
rchl: I think the path should be `images/tile-screenshots/qrcode.png`

styles/main.less
```diff
@@ -1105,6 +1106,13 @@ body {
    padding-top: 10px;
  }
 
+.item-image {
```
rchl: Since it's the same as `.item-camera`, you can add this class there instead of creating new block.

rchl:
Please remove newly added `scripts/.DS_Store`

TILE_EXAMPLES.md
christophebaraer: Good catch @rchl . I've renamed qrcode.png to IMAGE.png to comply with the existing naming convention.

styles/main.less
christophebaraer: Addressed.

scripts/controllers/main.js
christophebaraer: Addressed

christophebaraer:
Thanks for your review @rchl. All requested changes have been made.

styles/main.css
```diff
@@ -975,6 +978,12 @@ body {
 .item-gauge {
   padding-top: 10px;
 }
+.item-image {
+    height: 100%;
```
rchl: Please regenerate CSS file. This doesn't look like it came from scss compiler.

christophebaraer: I generated the CSS from the LESS file. Let me know if it needs additional adjustments.

rchl: It still looks weird, with extra blank lines added. I get different results with:

lessc --version                                                                                                                                                                                                               
lessc 3.11.1 (Less Compiler) [JavaScript]

rchl:
Pushed manually with some small style fixes. Thanks!

PR #336, opened 2020-01-27
Added darkmode option

srozemuller:
Added darkmode option in configfile.. true / false option
Added in the main.js an if statement to check the darkmode option and sets the correct class
Added in theme.css the darkmode colors.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -100,13 +100,19 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
          if(CONFIG.customTheme) {
             var themes = CONFIG.customTheme;
 
-            if(typeof themes === "string") themes = [themes];
-
-            themes.map(function (theme) {
-               bodyClass.push('-theme-' + theme);
-            });
+            if(typeof themes === "string") themes = [themes];		
+            	if(CONFIG.darkmode){
+            		bodyClass.splice(0, 1,'-theme-dark');
+               	}
+            	else{
+            		themes.map(function (theme) {
+	            		bodyClass.push('-theme-' + theme);
+	            	});
+	        }
+         
          }
-
+          
+          
          if(CONFIG.entitySize) {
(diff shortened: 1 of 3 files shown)


styles/themes.css
```diff
@@ -1,3 +1,165 @@
+:root{
```
rchl: You are supposed to modify *.less file and compile it to css, not modify css by hand.
Also, css variables are not supported by IE11 so not welcome in this project. But since LESS supports variables, you don't really need to use CSS variables.

scripts/controllers/main.js
```diff
@@ -100,13 +100,19 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+            if(typeof themes === "string") themes = [themes];		
+            	if(CONFIG.darkmode){
+            		bodyClass.splice(0, 1,'-theme-dark');
+               	}
+            	else{
+            		themes.map(function (theme) {
+	            		bodyClass.push('-theme-' + theme);
+	            	});
+	        }
+
```
rchl: Indentation is all messed up.

rchl:
I haven't tried the changes yet or looked into implementation but is the "darkmode" supposed to affect all themes? If so, have you tested all themes and made sure nothing looks weird or broken?

resoai:
I'm not sure introducing dark mode globally is such a good idea. First of all, this means that each and every theme needs to look OK which it never will, some themes are not meant to be dark by design. Why not create a separate dark theme and simply suffix its name with `-dark`?

srozemuller:
> I'm not sure introducing dark mode globally is such a good idea. First of all, this means that each and every theme needs to look OK which it never will, some themes are not meant to be dark by design. Why not create a separate dark theme and simply suffix its name with `-dark`?

styles/themes.css
srozemuller: I've modified the *.less file and compiled it to css.
For now the dark theme is based on the material theme.

PR #338, opened 2020-02-08
A different camera entity id and hls buffersize can now be specified

thingythingstuff:
A separate camera entity id can now be specified for fullscreen cameras, for example to show a high resolution rtsp stream on the popup, while the camera thumbnail tile uses a low resolution and low bandwidth still image sequence. If the id is not specified in the fullscreen object, then the tile item id is used by default, providing backwards compatibility.

The HLS buffer default length was reduced from 30 to 5 seconds to reduce latency, and can optionally be co figured from config.js. If bufferLength is not supplied, 5 seconds are used by default.

Usage:

`
fullscreen: {
   type: TYPES.CAMERA_STREAM,
   ObjFit: 'contain',
   id: 'camera.highres_camera',
   bufferLength: 3
}
`

### DIFF
--- a/TILE_EXAMPLES.md
+++ b/TILE_EXAMPLES.md
@@ -63,6 +63,31 @@ Manually trigger an automation
 }
 ```
 
+#### CAMERA_THUMBNAIL and CAMERA_STREAM
+Shows a camera feed on the tile and opens a fullscreen popup with an RTSP stream when pressed.
+Optionally, the fullscreen camera entity can be different from the thumbnail camera entity, for example to show a hires stream on the fullscreen popup only.
+
+```js
+{
+   position: [0, 0],
+   id: 'camera.front_gate',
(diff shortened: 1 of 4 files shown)


rchl:
The style (indentation, spaces within braces) is quite messy.
Also, we should add an example for CAMERA_STREAM in TILE_EXAMPLES.md as we shouldn't expect people to read source code to figure out what can be passed.

thingythingstuff:
The 'messy' white space between braces in conditional expressions is my code formatting style :) I also noticed that something was wrong with indentation after committing. I didn't think too much of it, as all that can be reformatted in seconds when merging. I guess I can correct it, if needed.

Agreed on the examples.

thingythingstuff:
Fixed whitespace stuff and added example.

rchl:
Thanks!

PR #340, opened 2020-02-13
Allow icon mapping on TILE.INPUT_SELECT

foosel:
By providing an `icons` array mapping possible values to
icons they will be displayed on the tile instead of the text value.

Example:

``` js
{
    position: [0, 1],
    type: TYPES.INPUT_SELECT,
    id: "input_select.house_mode",
    title: "House Mode",
    icons: {
        Normal: "mdi-home",
        Vacation: "mdi-palm-tree",
        Sick: "mdi-medical-bag",
        Travel: "mdi-airplane"
    },
}
```

image

Note that this is my first time interacting with Angular, so if there's a better way to achieve an "if-else" kinda construct in the template, I'm happy to learn about that and fix the code accordingly.


index.html
```diff
@@ -843,7 +843,11 @@
             </div>
          </div>
          <div class="item-triangle"></div>
-         <div class="item-entity -select">
+         <div ng-if="item.icons" class="item-entity">
+            <span ng-if="item.cons" class="item-entity--icon mdi"
```
rchl: `item.cons`? You meant `item.icons`? But it's not necessary to repeat that since it's on parent element already.

rchl:
Should be documented also.

index.html
foosel: Ups, sorry, yeah, that was a left-over from my experiments. Should be removed.

foosel:
> Should be documented in `TILE_EXAMPLES.md` also.

Will do so ASAP, almost done actually but image not showing up as it should, investigating.

foosel:
Fixed, squashed and pushed, ready for review. Thank you for considering!

PR #362, opened 2020-04-26
Use config serverUrl for HLS video stream

mathijsmeulendijks:
When serving TileBoard on a different url then HA API, HLS video streams are not working. 
Added the CONFIG.serverUrl to fix this issue

### DIFF
--- a/scripts/directives.js
+++ b/scripts/directives.js
@@ -219,8 +219,7 @@ App.directive('cameraStream', ['Api', function (Api) {
                },
                function (res) {
                   if(!res.result) return;
-
-                  appendVideo(res.result.url);
+                  appendVideo(CONFIG.serverUrl + '/'+ res.result.url);
                });
          };
          


resoai:
Thank you.

scripts/directives.js
```diff
@@ -219,8 +219,7 @@ App.directive('cameraStream', ['Api', function (Api) {
                },
                function (res) {
                   if(!res.result) return;
-
-                  appendVideo(res.result.url);
+                  appendVideo(CONFIG.serverUrl + '/'+ res.result.url);
```
rchl, line 222: Should probably use `toAbsoluteServerURL` here as this will break if URL is already absolute (whether it can be absolute, I don't know).

mathijsmeulendijks: Agree, `toAbsoluteServerURL` provides exactly what we need here, i'll send it in

PR #364, opened 2020-04-27
Fix for "Add to Home Screen" feature on iOS 13

danielfernau:
iOS 13 requires not only the `"apple-mobile-web-app-capable"` meta tag but also either one of the entries `"display": "standalone"` or `"display": "fullscreen"` in the manifest file. Otherwise the "Add to Home Screen" function only creates a link to TileBoard which opens the interface in Safari. With the manifest "display" entry it behaves more like a "progressive web app", hiding the browser UI and address bar.

https://developer.mozilla.org/en-US/docs/Web/Manifest/display

### DIFF
--- a/manifest.webmanifest
+++ b/manifest.webmanifest
@@ -4,5 +4,6 @@
     "src": "favicon.png",
     "sizes": "16x16 24x24 32x32 48x48 64x64",
     "type": "image/png"
-  }]
+  }],
+  "display": "fullscreen"
 }


jeremywillans:
Thanks for this PR @danielfernau - fixed this issue for me :)

Do you find that it does not refresh the content after closing/reopening the "App" ?

danielfernau:
Happy to hear that @jeremywillans :)
Yes, I experienced this issue a few times as well. What in most (but not all) cases works for me is to close it, swipe it out of the multitasking view and then re-open it. Not sure what's causing this, though... Probably has something to do with the way Safari handles site caches, tabs and (re)loading the web app in general.

PR #378, opened 2020-06-09
changed parameters to support more light types

hypoflex:
### index.html
thorough testing yesterday to why my light (with a max value of 254) did not appear at the correct value when long pressing.

apparently because `step, min and max` are being read AFTER `ng-change` on its first init the `max` is unknown and will be set to html's default of `100`.

setting `step, min and max` above `ng-change` with reading the value's for these directly from `slider` instead of the model fixed the issue.

changed `getLightSliderConf` variable to `_c_` instead of `_c` to better represent the code in `main.js`

### main.js
increased timer for opening the `long press` menu to give it more stability/accuracy.
added the check for `controlsEnabled` to make sure that `long press` does not do another `single press` after letting go.

### DIFF
--- a/index.html
+++ b/index.html
@@ -734,21 +734,21 @@
          <div ng-if="item.controlsEnabled" class="item-entity-sliders"
               ng-click="preventClick($event)">
             <div ng-repeat="slider in item.sliders track by $index"
-                 ng-if="(_c = getLightSliderConf(slider, entity))"
+                 ng-if="(_c_ = getLightSliderConf(slider, entity))"
                  class="item-slider-container">
 
                <div class="item-slider-title" ng-if="slider.title">
                   <span ng-bind="slider.title"></span>:
-                  <span ng-bind="getLightSliderValue(slider, _c)"></span>
+                  <span ng-bind="getLightSliderValue(slider, _c_)"></span>
                </div>
 
                <div class="item-slider">
-                  <input type="range" ng-model="_c.value" value="{{ _c.value }}"
+                  <input type="range" ng-model="_c_.value" value="{{ _c_.value }}"
                          ng-on-touchstart="$event.stopPropagation()"
                          ng-on-touchmove="$event.stopPropagation()"
                          ng-on-pointerdown="$event.stopPropagation()"
-                         ng-change="lightSliderChanged(slider, item, entity, _c)"
-                         step="{{ _c.step }}" min="{{ _c.min }}" max="{{ _c.max }}">
+						 step="{{ _c_.step }}" min="{{ _c_.min }}" max="{{ _c_.max }}"
+                         ng-change="lightSliderChanged(slider, item, entity, _c_)">
                </div>
             </div>
           
--- a/index.html
+++ b/index.html
@@ -747,7 +747,7 @@
                          ng-on-touchstart="$event.stopPropagation()"
                          ng-on-touchmove="$event.stopPropagation()"
                          ng-on-pointerdown="$event.stopPropagation()"
-						 step="{{ _c_.step }}" min="{{ _c_.min }}" max="{{ _c_.max }}"
+						 step="{{ slider.step }}" min="{{ slider.min }}" max="{{ slider.max }}"
                          ng-change="lightSliderChanged(slider, item, entity, _c_)">
                </div>
             </div>
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1014,6 +1014,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.toggleSwitch = function (item, entity, callback) {
+	  if(item.controlsEnabled) return;
       var domain = "homeassistant";
       var group = item.id.split('.')[0];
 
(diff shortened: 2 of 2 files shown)


Xitro01:
Got this running on my Tileboard instance and this fixes alot of issues that I was having with my lights tile.

index.html
```diff
@@ -734,21 +734,21 @@
...
                <div class="item-slider">
-                  <input type="range" ng-model="_c.value" value="{{ _c.value }}"
+                  <input type="range" ng-model="_c_.value" value="{{ _c_.value }}"
                          ng-on-touchstart="$event.stopPropagation()"
                          ng-on-touchmove="$event.stopPropagation()"
                          ng-on-pointerdown="$event.stopPropagation()"
-                         ng-change="lightSliderChanged(slider, item, entity, _c)"
-                         step="{{ _c.step }}" min="{{ _c.min }}" max="{{ _c.max }}">
+						 step="{{ slider.step }}" min="{{ slider.min }}" max="{{ slider.max }}"
+                         ng-change="lightSliderChanged(slider, item, entity, _c_)">
```
rchl, line 751: I saw your comment:
> apparently because step, min and max are being read AFTER ng-change on its first init the max is unknown and will be set to html's default of 100.
>
> setting step, min and max above ng-change with reading the value's for these directly from slider instead of the model fixed the issue.

But I can neither reproduce the problem nor really understand it.

The initial value for `step` values comes from slider definition so it should be that already. It will only fall back to entity attributes or 100 if those values are not defined in the slider.

Even if there is some problematic case, this changes the logic so that the value from the API will never be used when user didn't define his own in the slider. So that seems wrong.

Xitro01: The problem is that if you change it around and you define your own slider, it's not working either.
The problem I had is that it always defaulted to 100, which gave a lot of issues changing the brightness of my lamps (255 max).

rchl: I understand that there might be an issue but I need to in bestand and reproduce it so that it can be fixed properly. This seems more like a hack for specific use case that breaks other use cases.

rchl: I have both zwave and zigbee dimmable lights. They both work a bit different but I can't reproduce any issues like that in either.

I'd like to know which type of light are you using and the state dump from HA, both when it's on and off.

Xitro01: I have IKEA Zigbee dimmable RGB lights. This is from a lamp that's on, State: "on":
brightness: 228
hs_color:
  - 26.939
  - 38.43
rgb_color:
  - 255
  - 201
  - 157
xy_color:
  - 0.432
  - 0.368
friendly_name: Stalamp
supported_features: 49

This is from one that's off, State: "off":
friendly_name: Eettafel
supported_features: 49

rchl: Can you also post your tile configuration for that light?

Xitro01: Sure, they all have the same configuration:
`{
                            position: [2, 0],
                            title: 'Eettafel',
                            id: 'light.eettafel',
                            type: TYPES.LIGHT,
                            classes: ["-light-custom"],
                            states: {
                                on: "On",
                                off: "Off"
                            },
                            icons: {
                                on: "mdi-ceiling-light",
                                off: "mdi-ceiling-light",
                            },
                            sliders: [
                                {
                                    title: 'Brightness',
                                    field: 'brightness',
                                    max: 255,
                                    min: 0,
                                    step: 5,
                                    request: {
                                        type: "call_service",
                                        domain: "light",
                                        service: "turn_on",
                                        field: "brightness"
                                    }
                                }
                            ],
                            colorpicker: true
                        `},``

rchl: Cool. I also have ikea zigbee light and can reproduce so will look deeper into it when I have some time.

Xitro01: Awesome, thanks a lot! These changes fixed the issues for me, but would be great to fix this for other users (in all use cases) as well. I'm not a great programmer, but I'm willing to help where I can.

scripts/controllers/main.js
```diff
@@ -1014,6 +1014,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.toggleSwitch = function (item, entity, callback) {
+	  if(item.controlsEnabled) return;
```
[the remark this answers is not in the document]
hypoflex: understandable.
the reason for indentation was mainly for readability, the base code had so many weird indents specially with changes @Xitro01 introduced i couldn't stand reading it.

index.html
rchl: I'm fixing the issue with max value being clamped to 100 in https://github.com/resoai/TileBoard/pull/395.

scripts/controllers/main.js
```diff
@@ -869,7 +869,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
             $timeout(function () {
                item._controlsInited = true;
-            }, 50);
+            }, 500);
```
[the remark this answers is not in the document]
Xitro01: Hi @rchl this fixed either the long-press while the lamp was on or while the lamp was off. Did you try both scenario's and did both work properly?

PR #381, opened 2020-06-12
Weather Icon night state

twistp:
Weather Icon looks for a night icon based on sun.sun state.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -93,27 +93,27 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       }
    };
    
-   $scope.getCardinal = function(num) {
-       var val = Math.floor((num / 22.5) + 0.5);
-       var arr = [
-          {cardinal: "N", pointer:"↑"}, 
-          {cardinal: "NNE", pointer:"↗"}, 
-          {cardinal: "NE", pointer:"↗"}, 
-          {cardinal: "ENE", pointer:"↗"}, 
-          {cardinal: "E", pointer:"→"}, 
-          {cardinal: "ESE", pointer:"↘"}, 
-          {cardinal: "SE", pointer:"↘"}, 
-          {cardinal: "SSE", pointer:"↘"}, 
-          {cardinal: "S", pointer:"↓"}, 
-          {cardinal: "SSW", pointer:"↙"}, 
-          {cardinal: "SW", pointer:"↙"}, 
-          {cardinal: "WSW", pointer:"↙"}, 
-          {cardinal: "W", pointer:"←"}, 
-          {cardinal: "WNW", pointer:"↖"}, 
-          {cardinal: "NW", pointer:"↖"}, 
-          {cardinal: "NNW", pointer:"↖"}
+   var getWindDirection = function (degreeDirection) {
+       var value = Math.floor((degreeDirection / 22.5) + 0.5);
+       var direction = [
+          {windDirection: "N", pointer:"↑"}, 
+          {windDirection: "NNE", pointer:"↗"}, 
+          {windDirection: "NE", pointer:"↗"}, 
+          {windDirection: "ENE", pointer:"↗"}, 
+          {windDirection: "E", pointer:"→"}, 
+          {windDirection: "ESE", pointer:"↘"}, 
+          {windDirection: "SE", pointer:"↘"}, 
+          {windDirection: "SSE", pointer:"↘"}, 
+          {windDirection: "S", pointer:"↓"}, 
+          {windDirection: "SSW", pointer:"↙"}, 
+          {windDirection: "SW", pointer:"↙"}, 
+          {windDirection: "WSW", pointer:"↙"}, 
+          {windDirection: "W", pointer:"←"}, 
+          {windDirection: "WNW", pointer:"↖"}, 
+          {windDirection: "NW", pointer:"↖"}, 
+          {windDirection: "NNW", pointer:"↖"}
        ];
-       return arr[(val % 16)];
+       return direction[(value % 16)];
    }
    
    $scope.getBodyClass = function () {
@@ -656,7 +656,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
       if(!map){
          return (item.sun ? 
-                        $scope.getItemEntity({id: item.sun.id}).state == item.sun.night ? "nt-" : ""
+                        $scope.getItemEntity({id: item.sun.id}).state === item.sun.night ? "nt-" : ""
                         : "") + icon;
       }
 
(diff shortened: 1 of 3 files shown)


rchl:
It's this a generic fix that will work for all kinds of weather providers out only certain ones?

resoai:
One comment: `sun` components can be not enabled so we better check for the existence of it.

twistp:
Ok ... I will get this fixed

scripts/controllers/main.js
```diff
@@ -92,7 +92,30 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
          }
       }
    };
-
+   
+   $scope.getCardinal = function(num) {
```
rchl, line 96: 1) This doesn't have to be attached to the `$scope`. It can be just a normal function.
2) I was uncertain about the word "cardinal" as I didn't know the meaning of it so I've looked it up and I've learned that "Cardinal directions are the four main points of a compass: north, south, east, and west". Then there are also "ordinal directions" which refer to directions in-between of cardinal. And then there are also "secondary-intercardinal". So my point is that this naming is not really accurate. I'd suggest:
`function getDirection(???)`
I'm not sure what exactly the `num` signifies to be able to name it properly.
3) Generally, you give poor names to variables. `num`, `val`, `arr` means nothing. A good programming style suggests using meaningful names for variables and not using abbreviations.
4) This function can be moved close to the code using it (just after) rather than being somewhere completely else in the file.

scripts/controllers/main.js
```diff
@@ -661,7 +688,8 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    $scope.weatherListField = function (field, line, item, entity) {
       if(!line || !line[field]) return null;
 
-      return parseFieldValue(line[field], item, entity);
+      var retF = parseFieldValue(line[field], item, entity);
+      return (field == "cardinal") ? $scope.getCardinal(retF) : retF;
```
rchl, line 692: a) Use strict comparison `===`
b) naming again: `retF`?

scripts/controllers/main.js
```diff
@@ -631,7 +654,11 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
       if(typeof map === "function") return callFunction(map, [icon, item, entity]);
 
-      if(!map) return icon;
+      if(!map){
+         return (item.sun ? 
+                        $scope.getItemEntity({id: item.sun.id}).state == item.sun.night ? "nt-" : ""
```
rchl, line 659: Use strict comparison `===`

scripts/controllers/main.js
rchl, line 692: Are you missing some change in html file that would handle the `cardinal` field?
And per my other comment, not sure if `cardinal` is a good name for it. Unless maybe if Homeassistant uses that itself but I don't see such thing in yr integration at least.

rchl:
The additions need to be documented. At least in form of updating the examples in TILE_EXAMPLES.md.

twistp:
rchl - Thanks for all comments ... Will get all changed tomorrow. Will change function name to getWindDirection(windBearing){} as well as apply all other suggestions.

Cheers

scripts/controllers/main.js
twistp: Done

scripts/controllers/main.js
twistp: done

scripts/controllers/main.js
twistp: done

scripts/controllers/main.js
twistp: I changed the name to windDirection. The reason I am leaving this there is in case someone wants to return the wind direction. met.io passes the wind direction in degrees only. I am not sure about others but thought it would not hurt to add this functionality. I am using this eventually but I am not posting these changes yet. I have created a new Weather card with forecast and current weather conditions.

scripts/controllers/main.js
```diff
@@ -92,7 +92,30 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
          }
       }
    };
-
+   
+   var getWindDirection = function (degreeDirection) {
+       var value = Math.floor((degreeDirection / 22.5) + 0.5);
+       var direction = [
+          {windDirection: "N", pointer:"↑"},
```
resoai, line 99: Why do we even need repeating windDorection and pointer here? Why not {“N” :"↑"}?

twistp: The pointer is more of a graphical option. On my configuration I am displaying  ↑N 11.2 Km/h
for instance.

resoai: I meant in the code, there is no point to keep repeating those keys, simple key->value would do just fine.

scripts/controllers/main.js
rchl: > This function can be moved close to the code using it (just after) rather than being somewhere completely else in the file.

scripts/controllers/main.js
rchl: Can we keep PRs focused on a single thing?
It doesn't make sense to introduce something that is unused as I can't judge properly its implementation then.
The chances of someone using it are slim if it's not integrated in the core. First, someone has to find out that it exists which is hard in itself. Then someone would need to change the HTML which I presume not many are willing to do.

twistp: I can incorporate this to TYPES.WEATHER and TYPES.WEATHER_LIST if that makes more sense to you. I would just not be able to test it with dark_sky as I don't have it. Wind speed without direction is a bit useless. I am sure that other sailors out there will agree with me ;)
Let me know how to proceed.

What do you mean by keeping PRs focused on a single thing? All commits I made so far are related to weather and I thought it would be more logical to keep them in this single PR. Do you guys prefer that in another way?

scripts/controllers/main.js
twistp: It might be a stupid question but how would I display that value on the html if I don't know the key? The point about this function is that you pass degrees and the function spits out the direction value associate with that number. At this point all I have is a number.

scripts/controllers/main.js
rchl: > I can incorporate this to TYPES.WEATHER and TYPES.WEATHER_LIST if that makes more sense to you.

That would be better I think.

> I would just not be able to test it with dark_sky as I don't have it.

Dark sky is discontinued after Apple bought it so it needs to be removed from examples anyway.

> What do you mean by keeping PRs focused on a single thing?

Even though all commits are related to weather, that particular functionality is not used anywhere so I meant to introduce it together with changes that use it (your weather tile).

scripts/controllers/main.js
rchl: We don't understand the use case fully because of the thing I was talking about - there is no code that is using this functionality so we can't judge implementation properly.

This PR should be just about "weather icon night state" as you've called the PR yourself.

scripts/controllers/main.js
twistp: ok ... will make the changes and it will all make sense.

scripts/controllers/main.js
twistp: Apologies for that. I am new to github and still need to understand the PR concept. I will make the changes and hopefully the full commit will make sense. I will also make suggestions for the black sky (weather) implementation examples. And will do separate PRs for next time.

resoai:
I am closing this PR due to inactivity. Please feel free to open a new one taking into account new project structure.

PR #384, opened 2020-06-29
Add HVAC mode support to climate tile

resoai:
Unfortunately I have no way of testing it at all, please fell free to change anything you fell looks wrong.


resoai:
Thanks!

PR #387, opened 2020-07-02
Fix clicking menu on the left through the header

### LINKED ISSUE #385: The top icon cannot be clicked in the icon menu on the left.
If there are too many icons on the left, those that fall in the line of the header cannot be clicked. Thans for the fix! 
tileboard-issue

rchl:
Something we could do to address #385. At least for browsers that support pointer-events.

People would have to update their configs for this to be effective.

Not sure what implications changing from padding to margin can have on the header but seems to work in my layout at least.

Resolves #385


resoai:
Looks good!

PR #392, opened 2020-08-03
Implement Popup Tile

### LINKED ISSUE #388: New type of popup
A few users have asked for a popup which would be able to show a number of tiles. There is definitely something to it and we are already doing something similar with a door entry popup. How do you @rchl and @akloeckner feel about it? How would you call this type of popup?

akloeckner:
Proper (hopefully) minimal PR for modular popup. History will follow once merged.

Fixes #388


rchl:
It has some hvac stuff mixed in.

akloeckner:
That came from @resoai s dev branch. I did another rebase...

akloeckner:
> Something to keep note of: Tiles in the popup might themselves trigger popups. How to handle that?

Just a reminder that we're not fully done maybe...

rchl:
We need to document this in `TILE_EXAMPLES.md`.

rchl:
Created:
```js
            {
              width: 2, height: 2,
              type: TYPES.POPUP,
              id: 'light.corner',
            },
```
And getting this after opening the popup (clicking the tile):
Screenshot 2020-08-07 at 18 44 06

index.html
```diff
@@ -653,10 +670,16 @@
       <div ng-if="item.type === TYPES.POPUP_IFRAME"
            class="item-entity-container">
 
-         <div ng-if="item.customHtml"
-              ng-bind-html="itemCustomHtml(item, entity)"></div>
+           <div class="item-entity">
+            <span class="item-entity--icon mdi"
+                  ng-class="entityIcon(item, entity)"></span>
+         </div>
+      </div>
```
rchl, line 677: It seems like the code for `TYPES.POPUP_IFRAME` has changed. Was that on purpose?

resoai: This was still me removing customHtml part. I've no idea how it ended up there since only custom tile has it.

akloeckner:
> And getting this after opening the popup 

This seems to be a problem because you have no `popup` field in your test config. I could fix it by passing `item.popup || {}` to that function. I am just not sure how much of a failproofing like that you usually do in the project. I would bet something similar will happen with `item.popup.items` or the like...

rchl:
It doesn't necessarily have to be handled as long as there is documentation on how to configure the popup.

akloeckner:
> We need to document this

True, also in the readme! 

It will take me at least another week, until I have a decent computer at hand. Maybe @resoai can help out here. I'll take care of Read Me right away...

resoai:
I'm afraid I only have access via iPhone now.

akloeckner:
:-) Same here. So, we'll have to wait for examples and screenshots...

akloeckner:
> It doesn't necessarily have to be handled as long as there is documentation on how to configure the popup.

While we're waiting for those examples, I hope, the doc in readme is sufficient for further review.

resoai:
I'm sure it is! Those who understand the code will have a chance to test it, those who don't - won't even notice the difference.

akloeckner:
I found a way to add those examples...

TILE_EXAMPLES.md
```diff
@@ -399,6 +399,36 @@ Locks and unlocks the lock on tap. Also shows icon depending on the state<br>
...
+      tileSize: 100,
+      height: 3,
+      items: [
+         {
+            position: [1, 1],
+            type: TYPES.SENSOR,
+            id: 'sensor.covid',
+         },
+      ]
+   }
```
rchl, line 428: There is a lot of weird characters in this block (can't see on github).

index.html
rchl: So should be reverted?

rchl:
The sizing of the popup doesn't seem to work correctly. Even your screenshot shows the problem as I would expect the popup height to be no bigger than what's required to show the tile.

Also, if you change tile position to [1, 4] it will disappear outside of the popup (even if screen is big enough).

akloeckner:
Height was set to three in the example. So, it overrides counting the tiles. Should be the same behaviour as for groups, I believe.

TILE_EXAMPLES.md
akloeckner: Really weird. I just copied that from SSH to github. But, admittedly, that usually gives me the creeps with indentation. Always. No idea why. 

I fixed that.

index.html
akloeckner: I reverted it. People might be using this piece of code. @resoai, if you decide otherwise, re-reverting that single commit will be easy.

resoai:
@akloeckner I'm sorry to interrupt you guys and comment from mobile, but what is 30px you are subtracting in a few places? If this is size of the fixed element, perhaps this should be in a variable, like height of the title bar?

TILE_EXAMPLES.md
```diff
@@ -399,6 +399,36 @@ Locks and unlocks the lock on tap. Also shows icon depending on the state<br>
...
 ```
 
+#### POPUP
+Opens a popup when clicked. 
+The popup can be configured with any tile you like.
+Think of the popup as a single stand-alone `group` of tiles.
+<br>
+![SCRIPT](images/tile-screenshots/POPUP.png)
+<br>
+![SCRIPT](images/tile-screenshots/POPUP_popup.png)
```
rchl, line 409: - `SCRIPT` should be `POPUP`
 - It would be nice to have same size for the tile as existing ones (POPUP_popup will obviously be bigger)

TILE_EXAMPLES.md
```diff
@@ -399,6 +399,36 @@ Locks and unlocks the lock on tap. Also shows icon depending on the state<br>
...
+```js
+{
+   type: TYPES.POPUP,
+   id: {},
+   icon: 'mdi-android',
+   title: 'History popup',
+   state: false,
+   popup: {
+      tileSize: 100,
+      height: 3,
```
rchl, line 420: Can we maybe remove `height` from the example? It makes the popup look broken and people will be copy-pasting this.

akloeckner:
> what is 30px you are subtracting in a few places?

I believe it is the margin of the popup container. A variable would be better, yes!

TILE_EXAMPLES.md
akloeckner: We can, but it's quite a hassle. What is the usual size?

rchl: What do you mean that it's a hassle?

akloeckner: I have to change settings via SSH in my mobile, access the interface through port forwardings, crop screenshots, upload them to github. Everything's somewhat more complicated from the holidays. ;-)

rchl: I could fix some stuff like that if I could push to your branch. I believe there is a setting for that you can toggle.

akloeckner: That would  be very helpful. I'd appreciate! 

I have already set `allow edits from maintainers`. Maybe @resoai would need to make you one?

resoai: Believe that @rchl has all the powers in the world for this branch.

rchl: I can't get it to work.
I have added remote https://github.com/akloeckner/TileBoard.git/ but I can only fetch from it. On pushing I get authentication dialog but entering my credentials still results in an authentication error.

It could be that I just don't how to handle repos over HTTPS. For my own, I'm using SSH but I guess that doesn't work for remote ones.

rchl: OK, found out how to make HTTPS work using personal token but on pushing I'm getting:
```
To https://github.com/akloeckner/TileBoard.git
 ! [remote rejected] akloeckner-popup -> akloeckner-popup (permission denied)
```

resoai:
> > what is 30px you are subtracting in a few places?
> 
> I believe it is the margin of the popup container (2*15px). A variable would be better, yes!
> 
> Afterthought: I had the impression that variables would be better in various places (e.g., border around all popups, margin inside all popups). They could help making sure the look-and-feel is always consistent. Maybe, some more classes would even help users manage their custom.css. Should we change all that in a separate clean-up PR?
> 
> (I think, I'll still try and add this variable here at least.)

Yes, this is a good idea!

TILE_EXAMPLES.md
rchl: Addressed

TILE_EXAMPLES.md
rchl: Addressed.

README.md
```diff
@@ -378,6 +378,14 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
       backgroundColor: 'rgba(0, 0, 0, 0.1)', // Defaults to rgba(0, 0, 0, 0.1)
       fractionSize: 0 // Number of decimal places to round the number to. Defaults to current locale formatting
    },
+   /** type: POPUP **/
+   /* popup: object containing the popup definition.
+      The popup is basically a single page with a single group. */
+   popup: { // can contain page or group settings, e.g.:
```
rchl, line 384: Actually I'm thinking it should be called `group`, which it is.

rchl: Well, maybe not as it's not exactly a group. It's a mix of page and group...

resoai: In door entry popup it is still called `layout` which although is not perfect, is probably universal. I'm not saying it is right though.

rchl: I give up. I don't think that either DOOR_ENTRY solution would be good here nor I can think of a better one.

PR #394, opened 2020-08-11
History tile

akloeckner:
Adds a tile for displaying history of states. Just like the history popup, but in a tile.

See previous discussion also here: #388


akloeckner:
Question to the maintainers: does the z-index on .tile have a side-effect, I am not aware of?

scripts/controllers/main.js
```diff
@@ -1669,8 +1670,28 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+   $scope.initTileHistory = function (item, entity) {
+      var key = "_historyObject";
+
+      if(item[key]) return item[key];
+
+      item[key] = $scope.getHistoryObject(item, entity, item);
+
+      $timeout(function () {
+         item._historyInited = true;
+      }, 50);
```
rchl, line 1688: Unused?

index.html
```diff
@@ -1289,6 +1289,22 @@
...
                  ng-style="{'background-image': 'url(' + itemURL(item, entity) + ')'}"></div>
       </div>
+
+      <div ng-if="item.type === TYPES.HISTORY && (_h = initTileHistory(item, entity))"
+           class="item-entity-container">
+            <div class="item-history">
+               <div class="item-history--placeholder">
+                  <span ng-if="_h.isLoading && !_h.errorText">Loading history data...</span>
+                  <span ng-if="_h.errorText">{{ activeHistory.errorText }}</span>
+               </div>
```
rchl, line 1299: This text withing placeholder should be centered vertically as now it overlaps the `state` potentially.

index.html
```diff
@@ -1289,6 +1289,22 @@
...
+               <canvas id="item-history--canvas"
+                  class="chart chart-line"
+                  chart-data="_h.data"
+                  chart-dataset-override="_h.datasetOverride"
+                  chart-options="_h.options"></canvas>
```
rchl, line 1304: Should there be auto-update of history? Otherwise the state will get outdated quite quickly (depending on the use case of course).

TILE_EXAMPLES.md
```diff
@@ -223,6 +223,23 @@ Allows you to toggle a fan on and off, as well as set the speed.
...
+```js
+{
+   position: [0, 1],
+   width: 2, height: 2,
+   type: TYPES.HISTORY,
+   id: 'sensor.covid',
+   title: 'COVID-19 in Köln',
+   state: false,
+   offset: 5*24*3600*1000,
+   options: {legend: {display: false}}
```
rchl, line 239: Maybe we should disable legend by default for history tiles since there is only one entity displayed and the "title" should describe it already?

index.html
akloeckner: Good idea! I somehow thought, that would auto-magically be the case already. But it seems not so. What would I have to add to make that work?

Maybe we should add an option to disable update. Also the prominent animation should be disabled to allow frequent updates. What do you think?

TILE_EXAMPLES.md
akloeckner: Also history popups show one state only by default. Let's turn the legend off for single state histories by default, and default-enable it for more states. Ok?

scripts/controllers/main.js
akloeckner: Possibly. I copied that from a similar piece of code, where also I didn't really understand why it was there. I'll remove and see what happens...

akloeckner:
> It's probably hard to tell but can you explain why is that needed exactly?

I now remember that was part of the history popup using the new tile: If you have a first tile to fill the whole popup and a second tile to display somewhere on top of that, the chart would be drawn on top of the second tile. z-index fixed that. But it's not needed for the present PR, I think. So, I shall remove it for the time being, and handle later...

akloeckner:
Another thing I just noted: The history tile is missing the tooltip. I have no idea why. Do you guys have a clue I could check out?

scripts/controllers/main.js
rchl: It's not referenced anywhere else in the code so I'd say unused.

rchl:
> Another thing I just noted: The history tile is missing the tooltip. I have no idea why. Do you guys have a clue I could check out?

Because there is an element with class `.item-history--placeholder` on "top" of it. I guess we can just remove the placholder once history is loaded.

TILE_EXAMPLES.md
rchl: Disable also for the history popup you mean? I guess it's mostly fine, the only small thing is that the value **unit** is only shown in the legend but that is probably not a big deal since it should be clear typically.

index.html
rchl: - An option to disable update wouldn't hurt.
- Agree with disabling animation. And the placeholder text that tells that history is loading should only show on initial reveal, not on every update.
- As for updating the tile, hmm, I'm not sure from the top of my head. I'm sure there is an easy way that involves updating all tiles but we should avoid that. I can think of having like a map for the history tiles and then it would be easier to update specific objects within that map but that might require bigger refactorings. Maybe just leave that out for now and we can look in another PR.

scripts/controllers/main.js
akloeckner: Deleted with no noticed side effects.

index.html
akloeckner: Done. Centered also horizontally.

index.html
akloeckner: Simply updating chart.data on new state does the trick. It appears so lightweight that I decided to omit the option to disable updates. Only downside: time might not be 100% accurate on the chart for update data.

TILE_EXAMPLES.md
akloeckner: Done

akloeckner:
Removed z-index and placeholder (the latter only conditionally, obviously).

scripts/controllers/main.js
```diff
@@ -1668,9 +1669,57 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+            }, $scope.itemField('options', config, entity));
          });
+
+      if (typeof entityId === 'string') {
+         $scope.$watch(
+            function () {
+               return $scope.states[entityId].state;
+            },
+            function (newValue) {
+               historyObject.data[0].push({
```
rchl, line 1688: This is gonna fail on the initial run when `data` is not set yet.

scripts/controllers/main.js
```diff
@@ -1668,9 +1669,57 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+                 duration: 0
+               },
+               legend: {
+                 display: typeof entityId !== 'string'
+               },
+            }, $scope.itemField('options', config, entity));
          });
+
+      if (typeof entityId === 'string') {
+         $scope.$watch(
```
rchl, line 1683: Those watchers need to be unregistered or else those will accumulate when opening and closing popups.

scripts/controllers/main.js
```diff
@@ -1668,9 +1669,57 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+                     y: newValue
+                  });
+               });
+         });
+      }
+
+      return historyObject;
+   };
+
+   $scope.initTileHistory = function (item, entity) {
```
rchl, line 1711: It's not really an "init" as it's called from the template so it will be called a lot of times, depending on when angularjs feel like recreating the template. So that will create a lot of redundant network requests and also scope watchers.

rchl, line 1711: In general, I think the history shown in the tile should use quite different style from full graphs since there is little space in there and that makes them busy looking and not very readable.

For example, this is how my router's download speed is being shown:
Screenshot 2020-08-16 at 12 50 27

And here is how HA shows it (in a bigger tile but still):
Screenshot 2020-08-16 at 12 51 35

(HA is also showing 24h period so I guess it's smoothing the graph a lot in this case)

akloeckner: In my opinion, the HA example is pretty (but) useless. I can't read anything off a graph with so much smoothing and no axes. I'd say, a graph will just need to be given sufficient space...

Anyways, I now stared at your example for some time and played around with mine. Result is, I can't think of a decent (and universal) way to reduce space in a 1x1 tile while keeping the chart meaningful. So, I would set “okay“ defaults and leave optimization up to the user. Maybe, we could highlight some relevant options in the documentation. Suitable candidates:
* text format (in your case e.g. `8k`)
* wider tile
* padding around graph (also depends on whether title is shown or not)
* mirror ticks to be shown inside graph area (will likely require different axes color to look good)
* disable axes altogether
* reduce number of ticks shown
* ...

akloeckner: I could rename the function, sure. It's just similarly named as other functions of that sort.

The network requests and watchers should be created just once. Isn't that what storing the object in `item` is meant to prevent?

scripts/controllers/main.js
akloeckner: Good point, if people keep their board open for a long time, opening history popups a lot. It should go away, as soon as we migrate the history popup to use the tile, because then the history object (with its watchers) will be created only once and then stored in the config object.

It's still a good thing to provide a clean-up method. I'll do just that and add a `destroy` step in `closeHistoryPopup`...

scripts/controllers/main.js
akloeckner: Isn't `data` set just a few lines above?

rchl: No, it's set from an async callback triggered when API call returns. That happens later that this synchronous code.

scripts/controllers/main.js
rchl: Sorry, my bad. The method is called multiple times but storing the object in `item` prevents multiple requests. So all good.

scripts/controllers/main.js
akloeckner: I see. So, it's probably the most straight forward solution to simply create the watchers also in the callback, once the data is there. I'll fix that together with the destruction code.

scripts/controllers/main.js
akloeckner: I added a variable to `app.js` with much more suitable options for 1x1 tiles than before. Using this variable is simple for the user. And not using it by default puts the user in control. (Also, not using it by default will enable to use the history tile in a fullsize popup.)

scripts/controllers/main.js
```diff
@@ -1656,7 +1656,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
                }
 
                datasetOverride.push({
-                  label: seriesUnit ? (seriesName + ' / ' + seriesUnit) : seriesName,
+                  label: [undefined, '1', '', ' '].includes(seriesUnit) ? seriesName : (seriesName + ' / ' + seriesUnit),
```
akloeckner, line 1659: This is a simple fix we can include here as a convenience. But I can also set up a separate PR for it...

rchl: I'd prefer something like:
```js
seriesUnit.trim() && seriesUnit !== '1' ? (seriesName + ' / ' + seriesUnit) : seriesName
```

But I would really want to know why is there a `1` unit sometimes...

TILE_EXAMPLES.md
```diff
@@ -223,6 +223,29 @@ Allows you to toggle a fan on and off, as well as set the speed.
...
+   id: 'sensor.covid',
+   title: 'COVID-19 in Köln',
+   subtitle: function (item, entity) {
+      return 'since ' + timeAgo(Date.now() - (item.offset || 24*3600*1000));
+   },
+   offset: 5*24*3600*1000,
+   options: angular.merge(
+      MINIMAL_CHART_OPTIONS,
+      {layout: {padding: {bottom: 45}}}
+   )
```
rchl, line 245: I feel like this should be automatic. If there is subtitle then we should reserve less space for the chart.

rchl:
Great work so far :)

TILE_EXAMPLES.md
akloeckner: I'd support that. But I couldn't find a way to make that work seamlessly. Can you suggest a swift way to know in `getHistoryObject`:
* whether we're in a tile or in a popup
* how much space we have available, i.e.,
  * if there is a subtitle
  * which theme (and thus sizing of the title elements) we're set to?

scripts/controllers/main.js
akloeckner: I like your code. My experience is as follows:
* HA shows categorial plots, if `!seriesUnit`, i.e. `undefined` and even `''`.
* We thus need a non-empty unit for HA to show an actual graph.
* The "nice looking" way to solve this is setting the unit to `' '`. (This minimizes the unit view in HA badges.)
* The actual physical way to solve this is to set the unit to `1`. Because it's actually a simple count.
* Besides: I could imagine people using other ways to solve this (`-`, `/`, `~`, ...). 

So maybe, it's good to have the initial guess you suggest. And additionally supply an option to switch the legend manually. But then, it will become a complexer change to be put in a separate PR.

Let's stash this until October. ;-)

TILE_EXAMPLES.md
rchl: I'd have to look into it when I have some time.

scripts/controllers/main.js
akloeckner: Resolving this, as discussion continues here: https://github.com/resoai/TileBoard/pull/394#discussion_r473678477

TILE_EXAMPLES.md
rchl: It should be doable by placing a copy of the `item-subtitle` below the canvas element and setting it's text to `&nbsp` for example. Then it can be measured dynamically how far away the top of the element is relative to the parent's bottom edge.
But it seems like a fair amount of work so I won't block you on that.

akloeckner: Copying the elements was a great hint. I actually managed to do this without measuring anything. But instead I got acquainted with flex boxes...

akloeckner:
> 
> 
> Great work so far :)

Thanks! You're a hard reviewer as usual. ;-) It makes the result much better.

PR #395, opened 2020-08-15
Fix the issue with sliders max value being clamped to 100

rchl:


rchl:
The issue is in angularjs where it sets the model value before setting the `max` attribute which ends up with native range input being clamped to default max value 100.

PR #398, opened 2020-08-15
Avoid long tile titles wrapping and being clipped

rchl:


rchl:
To avoid cases like this:
Screenshot 2020-08-15 at 22 41 36

PR #399, opened 2020-08-16
Include unit in "state" field by default

rchl:


resoai:
I think that this might break functionality for current users.

rchl:
Can you think of specific example where it would? This only affects the case when user doesn't define the`state` property on a tile. The behavior will change in that case but for the better hopefully.

resoai:
Yes, I've checked our example climate tile and uses filter so it should not break anything. 

https://github.com/resoai/TileBoard/blob/master/TILE_EXAMPLES.md#climate

PR #406, opened 2020-08-23
Fix backwards compatibility with openPopup

### LINKED ISSUE #405: popup closing immediately

#404

rchl:
Resolves #405


akloeckner:
This will not fix #405. But it's a good fix.

PR #407, opened 2020-08-24
exchange `ng-click` for `hm-tap`

### LINKED ISSUE #405: popup closing immediately

akloeckner:
apparently ng-click receives ghost clicks, whereas hm-tap handles them fine.

fixes #405

as this touches almost all elements, maybe we should do some testing... I'll have it in my setup and report. So far: No problems.

### DIFF
--- a/index.html
+++ b/index.html
@@ -82,7 +82,7 @@
 
       <div class="camera-popup-container">
          <div class="camera-popup-title">
-            <div class="camera-popup-close" ng-click="closeCamera()">
+            <div class="camera-popup-close" hm-tap="closeCamera()">
                <i class="mdi mdi-close"></i>
             </div>
             {{ entityTitle(activeCamera, entity) }}
@@ -90,7 +90,7 @@
 
          <div class="camera-popup--list">
             <div class="camera-popup--list-item"
-                 ng-click="openCamera(item)"
+                 hm-tap="openCamera(item)"
                  ng-class="{'-active': activeCamera === item}"
                  ng-repeat="item in getCameraList() track by $index">
                <span ng-show="entity = getCameraEntityFullscreen(item)">
@@ -123,7 +123,7 @@
         ng-if="activeIframe && (entity = getItemEntity(activeIframe))">
       <div class="iframe-popup-container" ng-style="getPopupIframeStyles()">
          <div class="iframe-popup-title">
-            <div class="iframe-popup-close" ng-click="closePopupIframe()">
+            <div class="iframe-popup-close" hm-tap="closePopupIframe()">
                <i class="mdi mdi-close"></i>
             </div>
             {{ entityTitle(activeIframe, entity) }}
@@ -141,7 +141,7 @@
 
       <div class="door-entry-popup-container">
          <div class="door-entry-popup-title">
-            <div class="door-entry-popup-close" ng-click="closeDoorEntry()">
+            <div class="door-entry-popup-close" hm-tap="closeDoorEntry()">
                <i class="mdi mdi-close"></i>
             </div>
             {{ entityTitle(activeDoorEntry, entity) }}
@@ -174,11 +174,11 @@
 
    <div class="popup"
         ng-if="activePopup && (page = activePopup.layout || {})"
-        ng-click="closePopup()">
+        hm-tap="closePopup()">
 
-      <div class="popup-container" ng-click="preventClick($event)">
+      <div class="popup-container" hm-tap="preventClick($event)">
          <div class="popup-title">
-            <div class="popup-close" ng-click="closePopup()">
+            <div class="popup-close" hm-tap="closePopup()">
                <i class="mdi mdi-close"></i>
             </div>
             {{ entityTitle(activePopup.item, activePopup.entity) }}
@@ -192,12 +192,12 @@
    <div class="alarm-popup"
         ng-if="activeAlarm && (alarmEntity = getItemEntity(activeAlarm))">
 
-      <div class="alarm-popup-overlay" ng-click="closeAlarm()"></div>
+      <div class="alarm-popup-overlay" hm-tap="closeAlarm()"></div>
 
       <div class="alarm-popup-container"
            ng-class="{'-no-code': !alarmEntity.attributes.code_format}">
 
-         <div class="alarm-popup-close" ng-click="closeAlarm()">
+         <div class="alarm-popup-close" hm-tap="closeAlarm()">
             <i class="mdi mdi-close"></i>
          </div>
 
@@ -221,13 +221,13 @@
                <div class="alarm-popup-buttons-line"
                     ng-repeat="line in [6, 3, 0] track by $index">
                   <div class="alarm-popup-button" ng-bind="button + line"
-                       ng-click="inputAlarm(button + line)"
+                       hm-tap="inputAlarm(button + line)"
                        ng-repeat="button in [1, 2, 3] track by $index"></div>
                </div>
 
                <div class="alarm-popup-buttons-line">
-                  <div class="alarm-popup-button -l2" ng-click="inputAlarm(0)">0</div>
-                  <div class="alarm-popup-button -warning" ng-click="clearAlarm()">
+                  <div class="alarm-popup-button -l2" hm-tap="inputAlarm(0)">0</div>
+                  <div class="alarm-popup-button -warning" hm-tap="clearAlarm()">
                      <div class="mdi mdi-close"></div>
                   </div>
                </div>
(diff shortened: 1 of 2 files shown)


dtx360:
Did some testing, it broke the back button on longpress light sliders. If used inside popup-tile, the same back button closes the popup aswell. Other than that it seems to work fine on my devices on ordinary tiles.

akloeckner:
Hmm... the log says `$event.stopPropagation is not a function`. The same happens with elements using `preventClick`. Apparently, the event in hammer objects looks different from the angular events...

index.html
```diff
@@ -174,11 +174,11 @@
 
    <div class="popup"
         ng-if="activePopup && (page = activePopup.layout || {})"
-        ng-click="closePopup()">
+        hm-tap="closePopup()">
 
-      <div class="popup-container" ng-click="preventClick($event)">
+      <div class="popup-container" hm-tap="preventClick($event)">
```
akloeckner, line 179: `preventClick` needs to be fixed.

index.html
```diff
@@ -611,7 +611,7 @@
 
          <div class="item-fan">
             <div class="item-fan--speed" ng-if="entity.attributes.speed_list"
-                 ng-click="openFanSpeedSelect($event, item)">
+                 hm-tap="openFanSpeedSelect($event, item)">
```
akloeckner, line 614: `openFanSpeedSelect` needs to be fixed,

index.html
```diff
@@ -622,7 +622,7 @@
             <div class="item-select--option"
                  ng-repeat="option in entity.attributes.speed_list track by $index"
                  ng-class="{'-active': option === entity.attributes.speed}"
-                 ng-click="setFanSpeed($event, item, entity, option)">
+                 hm-tap="setFanSpeed($event, item, entity, option)">
```
akloeckner, line 625: `setFanSpeed` needs to be fixed

index.html
```diff
@@ -712,12 +712,12 @@
             <div ng-if="entity.state !== 'off'">
                <div ng-if="supportsFeature(FEATURES.LIGHT.BRIGHTNESS, entity)"
                     class="item-button -center-right"
-                    ng-click="increaseBrightness($event, item, entity)">
+                    hm-tap="increaseBrightness($event, item, entity)">
```
akloeckner, line 715: `increaseBrightness` and `decreseBrightnes` need to be fixed.

index.html
```diff
@@ -763,7 +763,7 @@
             </div>
 
             <div class="item-entity--back-button"
-                 ng-click="closeLightSliders($event, item, entity)">
+                 hm-tap="closeLightSliders($event, item, entity)">
```
akloeckner, line 766: `closeLightSliders` needs to be fixed.

index.html
```diff
@@ -773,11 +773,11 @@
 
          <div ng-if="entityState(item, entity) && entity.state !== 'off'">
             <div class="item-button -center-right"
-                 ng-click="dimmerAction('plus', $event, item, entity)">
+                 hm-tap="dimmerAction('plus', $event, item, entity)">
```
akloeckner, line 776: `dimmerAction` needs to be fixed

index.html
```diff
@@ -809,11 +809,11 @@
            class="item-entity-container">
          <div>
             <div class="item-button -center-right"
-                 ng-click="increaseNumber($event, item, entity)">
+                 hm-tap="increaseNumber($event, item, entity)">
```
akloeckner, line 812: `increaseNumber` and `decreaseNumber` need to be fixed

index.html
```diff
@@ -834,7 +834,7 @@
             <div class="item-select--option"
                  ng-repeat="option in entity.attributes.options track by $index"
                  ng-class="{'-active': option === entity.state}"
-                 ng-click="setSelectOption($event, item, entity, option)">
+                 hm-tap="setSelectOption($event, item, entity, option)">
```
akloeckner, line 837: `setSelectOption` needs to be fixed

index.html
```diff
@@ -1072,12 +1072,12 @@
          <div>
             <div class="item-button -center-right"
                  ng-if="entity.attributes.temperature && entity.state !== 'off'"
-                 ng-click="increaseClimateTemp($event, item, entity)">
+                 hm-tap="increaseClimateTemp($event, item, entity)">
```
akloeckner, line 1075: `increaseClimateTemp` and `decreaseClimateTemp` need fixing

index.html
```diff
@@ -1100,7 +1100,7 @@
             <div class="item-select--option"
                  ng-repeat="option in _options track by $index"
                  ng-class="{'-active': option === entity.state}"
-                 ng-click="setClimateOption($event, item, entity, option)">
+                 hm-tap="setClimateOption($event, item, entity, option)">
```
akloeckner, line 1103: `setClimateOption` needs fixing

index.html
```diff
@@ -1231,7 +1231,7 @@
             <div class="item-select--option"
                  ng-repeat="option in entity.attributes.source_list track by $index"
                  ng-class="{'-active': option === entity.attributes.source}"
-                 ng-click="setSourcePlayer($event, item, entity, option)">
+                 hm-tap="setSourcePlayer($event, item, entity, option)">
```
akloeckner, line 1234: Finally, `setSourcePlayer` needs to be fixed

akloeckner (commented):
There are some references to `$event` that need fixing. At least `preventClick` and `closeLightSliders` are confirmed to generate errors on `$event.stopPropagation`, which is not function.

akloeckner:
The undefined `stopPropagation` can be fixed with `$event.srcEvent.stopPropagation()` in `closeLightSliders`. But `preventClick` does not prevent clicks anymore, even with this fix.

Source for that piece of fix: https://github.com/RyanMullins/angular-hammer/blob/e42f4290931e39e10040a340d5665274e2446135/angular.hammer.js#L312

rchl:
I think this comment sums up my experience also: https://stackoverflow.com/a/46474373/986604

So basically to make this work, we need to call
```js
      $event.srcEvent.stopImmediatePropagation()
```
from `preventClick`.

No need for domEvent=true in that case.

index.html
akloeckner: `stopImmediatePropagation` seems to help here.

index.html
akloeckner: @dtx360, `stopImmediatePropagation` seems to help with sliders closing again. But the sliders do not reliably open on long-press, on my desktop. Do you know, if this is normal behavior? Also, on my mobile, the sliders open up again, when I click `back`. Do you see this, too?

rchl:
I haven't looked yet but it might be necessary to also test with browser that doesn't support pointer events as hammer uses Pointer events but fallback to touch events when not supported.

index.html
rchl: For me it seems to work as before or even better.
For zigbee light the sliders seems to show up consistently.
For zwave light the sliders consistently don't show up (like before, but before there was also a problem of turning off the light after releasing the button).

Tried on Desktop both in Desktop and Mobile mode.

scripts/controllers/main.js
```diff
@@ -1036,7 +1036,7 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
 
    $scope.dimmerAction = function (action, $event, item, entity) {
       $event.preventDefault();
-      $event.stopPropagation();
+      $event.srcEvent.stopImmediatePropagation();
```
rchl, line 1039: All those places can just call `$scope.preventClick` to avoid duplication.
And there is probably no need to return false anymore, when using hammerjs.

index.html
akloeckner: My first issue seems to be caused by a bunch of chained timeouts in openLightSliders. They seem to aim at ensuring the state to be on, when the sliders are shown. I'd suggest to remove all those timeouts and instead put another condition into ng-if. 

In this course, I'd also remove item._controlsInited, which is set through a very fast timeout of 50ms and only used when the slider is changed. I believe the user will take much more than 50ms to change the slider anways...

rchl: Feel free to do those changes and I'll try them out later.

scripts/controllers/main.js
akloeckner: Good point. done.

rchl:
With latest change the "back" button on sliders tile turns off the light.

akloeckner:
Interesting. It doesn't for me. Instead the sliders reappear as described. Only on mobile. Could it be that some touch-up event causes all those issues? I could imagine tap is fired on the sliders and prevented to propagate. And then we lift the finger on the regular tile. On Firefox mobile that's interpreted as a press, on your device as another tap, and on my laptop it's ignored? 

Anyways, I have the impression that we're running into quite some trouble here, if lights alone pose such challenges. Maybe, we should switch out hm-tap place by place. We could start with the popup, where things seem to work fine.

akloeckner:
I just noticed: the sliders also open, when I:
1. tap the light to turn it on
2. tap decrease brightness.

akloeckner:
When closing the popup by clicking outside its frame, even underlying tiles get activated. As I don't understand what is going on, I'm afraid I feel unable to fix those issues. I'll propose a different fix in a new PR.

akloeckner:
As said in #408 this was not the answer. We might need to bear through the pains here. But I'm at the end of my wisdom with how different ly things behave on different devices

rchl:
I can't believe that angualr-hammer is so broken...
For every element using hm-* directives it adds new event listeners both on the element and on the window. Those on the window are triggered in the order they were added so those will always trigger first on the tile before triggering on tile's child button. This seems so broken.

I think I'm gonna rewrite this whole thing in Vue or something as it's really annoying to work with those unmaintained libraries.

akloeckner:
:-) At least I am not the only one to be overchallenged with the twirks of hammer working with angular...

However, I am also no-one to give up too easily. That's why I gave it another try using the original hint on the hammer website. I found, it's no use to prevent *all* ghost clicks, because ng-click wouldn't receive *any* clicks anymore. So, I took parts of the code and added it in the TileBoard code to prevent only selective ghost clicks. Currently, only the ones generated after a secondary action. So, this is more like a minimal-invasive solution instead of the (apparent) nuke we tried here. ;-)

You'll find the new take here: 
https://github.com/akloeckner/TileBoard/tree/ghost2

If you find the time to take a look, I'd appreciate comments. I can also open another PR. But I don't want to spam those test PRs all over the place...

rchl:
I'd like to stay on this one a little longer to see if anything can be done in a more generic way.

I've noticed that the order in which directives are linked does not necessarily follow the logical DOM order so even if Angular Hammer would want to, it can handle things properly in regard to order. Or at least not if the order is tied to directive linking.

I'm trying out some fundamental changes in Hammer.js but it ain't gonna be a safe change most likely.

akloeckner:
Alright. I'm happy with my fix so far, though. So, if we don't find a generic solution, we can pull it as a backup. Let me know, when I can do anything further but wait.

rchl:
Back in the day I've spent way too much on this and was not able to come up with any sensible solution. So we'll go with your workaround in another PR probably.

PR #408, opened 2020-08-26
Fix ghost clicks on popup

### LINKED ISSUE #405: popup closing immediately

akloeckner:
Fixes #405 
Closes #407 

This is a different fix for issue #405 . It uses the same mechanism already in place for light controls. It puts the mechanism also to the light controls back button.

There is now a variable to set the timeout, because 50ms were not sufficient for my apparently slow device. The user should thus be able to adopt to their devices' capabilities. I found 500ms sufficient to cancel most ghosts. I believe it will not harm responsiveness too much. After all, humans will need roughly 1000ms to react to anything anyways...

### DIFF
--- a/scripts/app.js
+++ b/scripts/app.js
@@ -226,6 +226,8 @@ var MINIMAL_CHART_OPTIONS = {
    },
 };
 
+var GHOST_CLICK_TIMEOUT = 500;
+
 window.onerror = function (error, file, line, char) {
    var text = [
       error,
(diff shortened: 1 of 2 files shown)


akloeckner:
:-( 

Just noticed that this does fix ghosts on the popup closing areas. Obviously not so much for ghosts on actual tiles in the popup. We'll need a more general solution. Possibly well have to bear through the pains in #407...

PR #411, opened 2020-08-30
clean up popup-related styles and fix size issue

akloeckner:
My old iPad would be confused by the two definitions of popup-container and it would always display the history popup fullsize. Now it's less cluttered and unambiguous.

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -1079,6 +1079,7 @@ camera_stream div {
   box-sizing: border-box;
   background: rgba(45, 49, 53, 0.78);
 }
+<<<<<<< Updated upstream
 .camera-popup-container,
 .door-entry-popup-container,
 .iframe-popup-container,
@@ -1091,6 +1092,8 @@ camera_stream div {
   position: relative;
   margin: auto;
 }
+=======
+>>>>>>> Stashed changes
 .camera-popup-title,
 .door-entry-popup-title,
 .iframe-popup-title,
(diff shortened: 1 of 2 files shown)


styles/main.less
```diff
@@ -1345,8 +1345,10 @@ camera_stream {
    justify-content: center;
    align-items: center;
 
-   &-container {
+   .popup-container { 
       /* auto-size container */
+      // Not simply &-container, in order to make this more specific
+      // and override the less specific height and width settings above
```
rchl: Which ones do you mean by "above"?

But then, in the end, you have changed it back anyway.

styles/main.css
```diff
@@ -1079,6 +1079,7 @@ camera_stream div {
   box-sizing: border-box;
   background: rgba(45, 49, 53, 0.78);
 }
+<<<<<<< Updated upstream
```
rchl, line 1082: Conflict resolution has gone wrong.

rchl (changes requested):

styles/main.less
akloeckner: Yes, I let that change sink in for a while. Then figured it was not so clean.

styles/main.css
akloeckner: Ups. Fixed that.

rchl:
I've pushed some changes to have less duplication. The changes should be equivalent to yours otherwise. Let me know if this works for you.

akloeckner:
Yes, seems to work! I removed one other style you seem to have missed in re-using.

PR #412, opened 2020-08-31
Simplify HTML by combining all "simple" tiles into one template

akloeckner:
When setting up popups, I came across the fact that many many tiles actually had the very same content. I fixed that up into one tile template. Basically not much changes, but the code gets much smaller and simpler. 

In this course, I also removed one superfluous function, because the existent and more generic function `itemField` could be used instead.

Things that do change:
* all simple tiles have `customHtml` option now.
* the customHtml option additionally accepts templates now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    * You can use any of the material design icons from https://materialdesignicons.com/
    * as long as they have been published in the [latest npm package](https://github.com/templarian/materialdesign-svg).
    */
-  icon: 'mdi-phone'
+  icon: 'mdi-phone',
   /* customHtml: Replace the icon by a custom HTML content
    * If you define this option, simple tiles will not show the icon. They will show your custom HTML instead.
    * You can use an anonymous function here, or a string that will be interpreted as a template
--- a/README.md
+++ b/README.md
@@ -248,6 +248,12 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    * as long as they have been published in the [latest npm package](https://github.com/templarian/materialdesign-svg).
    */
   icon: 'mdi-phone'
+  /* customHtml: Replace the icon by a custom HTML content
+   * If you define this option, simple tiles will not show the icon. They will show your custom HTML instead.
+   * You can use an anonymous function here, or a string that will be interpreted as a template
+   */
+  customHtml: function (item, entity) {return '<b>' + entity.state + '</b>';},
+  customHtml: '<b>&group.lights.state</b>',
   /* bg: Link to a background image for the tile
    * @ and & prefixes are explained below
    */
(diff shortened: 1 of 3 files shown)


scripts/controllers/main.js
```diff
@@ -521,14 +521,6 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       return getItemFieldValue(field, item, entity);
    };
 
-   $scope.itemCustomHtml = function (item, entity) {
-      if (typeof item.customHtml === 'function') {
-         return callFunction(item.customHtml, [item, entity]);
-      }
-
-      return item.customHtml;
```
rchl, line 529: Going through `parseString` rather than just using the string verbatim like before can break existing use cases. I think we should keep previous behavior. At least by default.

README.md
```diff
@@ -248,6 +248,12 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    * as long as they have been published in the [latest npm package](https://github.com/templarian/materialdesign-svg).
    */
   icon: 'mdi-phone'
```
rchl, line 250: Let's add the missing comma here as otherwise highlighting breaks in my editor at least.

rchl (changes requested):

scripts/controllers/main.js
akloeckner: I thought about that, but I couldn't really think of a (probable) use case, which would break. It would only break in the rare case that someone uses a `@` (or `&`) followed by a verbatim HA attribute (or state) name, even including the `.`s. Why would someone want to have that in their custom HTML? 

The most probable would be `@state`. But even that is a string I'd judge uncommon to appear in a tile. 

On the other hand, I even found it counter-intuitive that in all other places, template can be used interchangeably with anonymous functions. But not here. I'd assume, it is even an oversight that it wasn't possible before. 

That's why I suggest to unify this. (Here and in most other places we can find, I'd dare say.)

But maybe you have reasonable use cases, I didn't think of. Or @resoai can recall the reason, this part should not accept templates? (The code was introduced in #318, just a month ago.)

Another aspect to keep in mind is that probably we cannot always maintain 100% backwards compatibility, if we want to improve things beyond a certain level. I'd judge consistency the higher value. If you keep compatibility up as a core value for the project, though, I'll certainly revert this change.

README.md
akloeckner: I already wondered, why the readme looked strange in that place... done.

scripts/controllers/main.js
resoai: It does indeed look like an oversight to me. New tiles have been added quickly, to be honest I never expected it to grow over 5 since I was planning for it to be used solely for my dashboard 😄

akloeckner: :-) Isn't it amazing how fast good ideas are picked up?

akloeckner: Another option how to handle incompatible CONFIG changes would be to show warnings for a while: https://github.com/akloeckner/TileBoard/commit/4ab21fc509200d35cf535aeb92b05c296d1f439d. After some time we could get rid of those warnings and *really* clean up. But code will become messy with those warnings...

rchl: Let's keep it simple and just hope nothing breaks.

PR #417, opened 2020-09-04
Add support for passing variables when running scripts

### LINKED ISSUE #416: Pull variables from type.script
How to directly pull variables from type.script
now i use  this.apiReques call_service to pull three variables:
                                type: 'call_service',
                                 service: 'turn_on',
	                         domain: 'script',
                                 service_data: { entity_id: "script.common_timer", variables:{"duration":"00:10","entity_id":"switch.mp1_mp11","operation":"switch.toggle"} }

rchl:
Resolves #416

Disclaimer: Haven't tested it.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1107,12 +1107,12 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.callScript = function (item, entity) {
-      var variables
+      var variables;
 
       if(typeof item.variables === "function") {
          variables = callFunction(item.variables, [item, entity]);
       } else {
-         variables = item.variables || {}
+         variables = item.variables || {};
       }
 
       sendItemData(item, {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1107,12 +1107,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.callScript = function (item, entity) {
+      var variables
+
+      if(typeof item.variables === "function") {
+         variables = callFunction(item.variables, [item, entity]);
+      } else {
+         variables = item.variables || {}
+      }
+
       sendItemData(item, {
          type: "call_service",
          domain: "script",
          service: "turn_on",
          service_data: {
-            entity_id: item.id
+            entity_id: item.id,
+            variables: variables
          }
       });
    };
--- a/README.md
+++ b/README.md
@@ -286,6 +286,12 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    */
   hidden: Boolean | Function,
   /*** TILE SPECIFIC SETTINGS ***/
+  /** type: SCRIPT **/
+  /* variables: Provides extra variables to the script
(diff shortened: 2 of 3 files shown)


scripts/controllers/main.js
```diff
@@ -1107,12 +1107,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.callScript = function (item, entity) {
+      var variables
```
akloeckner, line 1110: Missing `;`?

scripts/controllers/main.js
```diff
@@ -1107,12 +1107,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+
+      if(typeof item.variables === "function") {
+         variables = callFunction(item.variables, [item, entity]);
+      } else {
+         variables = item.variables || {}
```
akloeckner, line 1115: Missing `;`?

rchl:
> How about simply adding `data` which will be merged with service_data? Perhaps this could than also be used somewhere else.

Is there any other data that user could pass to the script? Because if not then IMO making this more generic would just make it less easy to use. User would have to understand whether he needs the variables object there or just define them as root keys.

scripts/controllers/main.js
rchl: I've stopped caring about style in this code base ;)
Can fix tomorrow as I'm not at the computer anymore.

akloeckner:
> Is there any other data that user could pass to the script?

Looking at the code, the first thing that comes to mind is `entity_id` as output of a function. But looking at https://www.home-assistant.io/integrations/script/#passing-variables-to-scripts, the `data` approach will also involve calling a different service, with no `entity_id` in the `data`.

Btw, I was suprised about the `variables` and would have expected `data`. But both are correct obviously...

rchl:
> Btw, I was suprised about the `variables` and would have expected `data`. But both are correct obviously...

I believe that using `script.turn_on` with particular entity_id requires passing a `variables` object while calling the `script.script_name` directly requires passing variables as plain `data`. But our `callScript()` uses the former so we need to pass the `variables` object.

> PS. Thinking about it... renaming callScript to callService, passing service (=script) id and data. Would allow to re-use for other tiles more generically, maybe?

I can do this refactor but I'd rather do it separately from this.

> How about simply adding `data` which will be merged with service_data? Perhaps this could than also be used somewhere else.

This will not work:

```js
{
   title: 'SCRIPT',
   id: 'script.test_script',
   type: TYPES.SCRIPT,
   data: {
      value: 'value from script'
   }
},
```

because the `script.turn_on` service requires that the variables are within the `variables` object.

And requiring the user to specify it like so:

```js
{
   title: 'SCRIPT',
   id: 'script.test_script',
   type: TYPES.SCRIPT,
   data: {
      variables: {
         value: 'value from script'
      }
   }
},
```

is a bit pointless IMO as there is nothing else that can be specified in the `data` itself (apart from `entity_id` which is taken from the `id` property automatically).

rchl:
@resoai what do you think about my last comment. Would you still want something to be changed or keep as is?

resoai:
@rchl since there is hardly any way this can be reused, I would be inclined to think that we could keep just variables as proposed.

PR #421, opened 2020-09-06
Create "callService" to replace "sendItemData" and have less repetition

rchl:
@akloeckner

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1835,12 +1835,12 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       var str = $scope.getActiveDatetimeInput();
       var dt = str.split(' ');
 
-      var data = {};
+      var serviceData = {};
 
-      if(entity.attributes.has_date) data.date = dt[0];
-      if(entity.attributes.has_time) data.time = dt[1] || dt[0];
+      if(entity.attributes.has_date) serviceData.date = dt[0];
+      if(entity.attributes.has_time) serviceData.time = dt[1] || dt[0];
 
-      callService(item, 'input_datetime', 'set_datetime', data)
+      callService(item, 'input_datetime', 'set_datetime', serviceData)
    };
 
    $scope.inputDatetime = function (num) {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1147,8 +1147,6 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    };
 
    $scope.setLightBrightness = function (item, brightness) {
-      if(item.loading) return;
-
       var serviceData = {
          brightness_pct: Math.round(brightness / 255 * 100 / 10) * 10
       };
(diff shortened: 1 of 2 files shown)


scripts/controllers/main.js
```diff
@@ -975,22 +967,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
...
+      setSliderValue(item, entity, value);
    };
 
    $scope.lightSliderChanged = function (slider, item, entity, value) {
       if(!item._controlsInited) return;
       if(!slider._sliderInited) return;
       if(!entity.attributes._sliderInited) return;
 
-      setSliderValue(item, entity, value, false);
+      setSliderValue(item, entity, value);
```
rchl, line 984: Those 4th arguments were not really used for anything.

scripts/controllers/main.js
```diff
@@ -1965,17 +1835,12 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       var str = $scope.getActiveDatetimeInput();
       var dt = str.split(' ');
 
-      var data = {entity_id: item.id};
+      var data = {};
```
resoai, line 1838: Perhaps this thinks also be named `serviceData`.

scripts/controllers/main.js
```diff
@@ -1212,15 +1149,11 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
    $scope.setLightBrightness = function (item, brightness) {
       if(item.loading) return;
```
akloeckner, line 1150: We could get rid of those types of lines as well, because it's in `callService` already.

scripts/controllers/main.js
```diff
@@ -2161,12 +2026,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       });
    }
 
-   function sendItemData (item, data, callback) {
+   function callService(item, domain, service, data, callback) {
       if(item.loading) return;
```
akloeckner, line 2030: Why is this stuff with `item.loading` here? This basically acts as rate limiter for service calls?

scripts/controllers/main.js
```diff
@@ -2161,12 +2026,21 @@ App.controller('Main', ['$scope', '$timeout', '$location', 'Api', function ($sco
       });
    }
 
-   function sendItemData (item, data, callback) {
+   function callService(item, domain, service, data, callback) {
```
akloeckner, line 2029: If `callService` was item-agnostic, we could make it a scope method, such that people could re-use it more freely. Or we could put it in `Api`. Or expose it in the context of anonymous functions.

akloeckner (commented):
I haven't had the opportunity to actually test this yet, but it looks good. Some comments are in the review.

scripts/controllers/main.js
rchl: Not really rate limiter but to prevent accidental double clicks triggering actions twice. `item.loading` is set back to `false` after API call resolves.

akloeckner: I see. So this makes sure the tile is up to date, before another command is sent. Protecting the user from re-clicking, “because nothing has happened“. Makes sense.

scripts/controllers/main.js
rchl: I've added it to the `Api`.

scripts/controllers/main.js
rchl: Done

akloeckner:
:+1: I still haven't been able to test. But most of the tiles are not in my setup anyway. So, don't wait for it...

PR #432, opened 2020-09-24
Use rollup to build the app

rchl:
It's now an npm project with those advantages:
- introduces a build step to allow for using modern code syntax
  that will be automatically transpiled to support older browsers.
- includes commit hook to automatically lint the code on committing
- enables (selectively) type checking to catch issues early during development
- allows to clean up the code and separate into smaller chunks that will
  be combined into a single bundle
- releases will be versioned so that it's possible to refer to concrete
  versions and possibly include changelogs.
- versioning also allows to have non-random URLs for included scripts
  so that those can be cached by the browser

The end-user code is now generated in dist/ so that is what user has to
deploy rather than all files in the repo.


rchl:
@resoai @akloeckner @cgarwood It's a pretty serious change. For those who are involved in the development or plan to be, I encourage you to try it and report issues, annoyances, or anything that might be wrong with this approach. We go forward with it only when everyone is happy. :)

rchl:
BTW. The approach with `document.write`'ing the config file from the index.html is neither great nor very compatible with this approach. It required to do it in sort of hacky way by processing globals/constants and the rest of the app in two input steps, and then `document.write`'ing the config in between.

A proper solution would be to make the config file be an angular service that could be loaded asynchronously but that would be a breaking change that would require all user configs to be adapted. So I don't really see this happening anytime soon.

rchl:
Also, I did as minimal as possible changes to existing scripts. If this gets accepted and merged we can do further things with the scripts like:
 - using more modern JS
 - enabling more limiting rules
 - structuring script files better
 - possibly splitting some code into smaller chunks (it all gets bundled together anyway)

rchl:
I'm sorry but I'll recreate this PR in this repo so that it's easier to check out and test.

Issue #425, opened 2020-09-15
Scripts with same id overwrite background 


tobycheese:
I have a bunch of tiles that call the same script, as follows:

               {
                  position: [0, 0],
                  type: TYPES.SCRIPT,
                  id: 'script.sonos_wallpanel',
                  bg: 'images/radiostations/radioparadise.png',
                  variables: {source: 'radioparadise'},
               },
               {
                  position: [1, 0],
                  type: TYPES.SCRIPT,
                  id: 'script.sonos_wallpanel',
                  bg: 'images/radiostations/demoscene.png',
                  variables: {source: 'demoscene'},
               },

Clicking these launch the script with different variable values, this works as expected. However, both tiles show the same background image radioparadise.png. When changing the id of the second tile, the background is then respected (demoscene.png is shown for the second tile).

akloeckner:
I believe, the background is cached in the state entity, and not in the tile item. Try changing that line to `var obj = item;`:

https://github.com/resoai/TileBoard/blob/538d00a40a7418f4058cd3eae795e676a0ade9e6/scripts/controllers/main.js#L415

Now, the question is why this is how it is. I'd suspect it is just a coincidence and most re-appearances of this piece of code could be changed in the way to use `item` instead of `entity`.

tobycheese:
@akloeckner Yes, changing that fixed the issue, thanks! Will this result in a fix eventually?

rchl:
Makes sense to change that IMO.

rchl:
Fix released in https://github.com/resoai/TileBoard/releases/tag/v2.0.2

Issue #386, opened 2020-07-02
History API CORS issue


resoai:
I've just moved my tileboard to a different server and noticed that all API calls to history are failing.

```
Access to XMLHttpRequest at 'http://192.168.1.2:8123/api/history/period/2020-07-01T11:22:05.828Z?end_time=2020-07-02T11:22:05.828Z&filter_entity_id=person.marina' from origin 'http://192.168.1.30' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

@akloeckner did you expirience anything similar?

rchl:
https://www.home-assistant.io/integrations/http/#cors_allowed_origins

resoai:
Totally forgot about it, however it doesn't look like it is working for API.

rchl:
Yes, I also can't seem to make it work...

I've tried with a "postman" kind of app and it worked that way (with token taken from TileBoard).

It seems to fail from Tileboard because the browser triggers OPTIONS request that HA doesn't seem to be handling properly (405 Method Not Allowed).

The OPTIONS request is made because request is "non-simple" due to having `Authorization` header.

resoai:
2018!

https://github.com/home-assistant/core/issues/13400

akloeckner:
I haven't seen such an error. But my Tileboard is served by HA's `www` folder. So, this issue wouldn't appear...

akloeckner:
To be more solution-oriented: it seems the issue linked above also has an easy fix. Maybe it's best to open a PR to HA's core with it? 

To be pessimistic again: my recent experience with a PR of mine leads me to think it will take some substantial amount of time until we get this merged...

resoai:
Keeping all files in www directory, especially if you hardcode the token is a big security risk. I've even added a warning to our readme file today. I rarely use charts anyway but it is something which would be nice to get fixed. I've tried creating a PR in the past and gave up :)

akloeckner:
My Tileboard is only available on my local network and the token is not hardcoded. So, I figured it should be ok to have everything in www. However, as it's not recommended, we should try to get a fix into HA...

akloeckner:
If I read this right, `OPTIONS` should be allowed as part of `methods=*`. What do you think? 

https://github.com/home-assistant/core/blob/4679e670f194931a083b693891fc802cd8857920/homeassistant/components/http/cors.py#L32

resoai:
[...]

PR #433, opened 2020-09-24
Use rollup to build the app

rchl:
It's now an npm project with those advantages:
- introduces a build step to allow for using modern code syntax
  that will be automatically transpiled to support older browsers.
- includes commit hook to automatically lint the code on committing
- enables (selectively) type checking to catch issues early during development
- allows to clean up the code and separate into smaller chunks that will
  be combined into a single bundle
- releases will be versioned so that it's possible to refer to concrete
  versions and possibly include changelogs.
- versioning also allows having URLs without random string attached
  so that those can be cached by the browser

~~The end-user code is now generated in dist/ so that is what user has to
deploy rather than all files in the repo.~~


rchl:
@resoai @akloeckner @cgarwood It's a pretty serious change. For those who are involved in the development or plan to be, I encourage you to try it and report issues, annoyances, or anything that might be wrong with this approach. We go forward with it only when everyone is happy. :)

<hr>

BTW. The approach with `document.write`'ing the config file from the index.html is neither great nor very compatible with this approach. It required to do it in a sort of hacky way by processing globals/constants and the rest of the app in two input steps, and then `document.write`'ing the config in between.

A proper solution would be to make the config file be an angular service that could be loaded asynchronously but that would be a breaking change that would require all user configs to be adapted. So I don't really see this happening anytime soon.

<hr>

Also, I did as minimal as possible changes to existing scripts. If this gets accepted and merged we can do further things with the scripts like:
 - using more modern JS
 - enabling more linting rules
 - structuring script files better
 - possibly splitting some code into smaller chunks (it all gets bundled together anyway)

akloeckner:
Wow. Looks like a great deal of changes. I'll have to deep dive into npm and test with some of my devices... I'll report back!

akloeckner:
So, I checkout out your branch and copied my config files to the dist folder. I receive the "invalid config" error... Is stuff like that not allowed anymore?

```
testEntity='sensor.aktuell_infizierte_in_koln';
//testEntity='input_select.mode';
//testEntity='input_number.wohnzimmer_ontime';

/**
 * Load JS file synchronously and evaluate
 */
function loadJS(url) {
   var xhttp = new XMLHttpRequest();
   var script = document.createElement( "script" );
   xhttp.open("GET", url+suffix, false);
   xhttp.send();
   script.text = xhttp.responseText;
   document.head.appendChild( script ).parentNode.removeChild( script );
}
loadJS('autoconfig.js');

/**
 * AUTOMATIC TILE SIZING
 *
 * Calculates tile size automatically for different screensizes.
 */

var groupWidth  =   2; // count of tiles horizontally
var groupHeight =   4; // count of tiles vertically
var tileMargin  =   6; // px between tiles
var groupMargin =  50; // px between groups
var headerSize  =  60; // px of group or page header
var maxTileSize = 120; // px

var tileSize = setTileSize();
window.addEventListener('resize', function () {CONFIG.tileSize = setTileSize();});

function setTileSize () {
   var width   = window.innerWidth  - headerSize - groupMargin - tileMargin * (groupWidth  - 1);
   var height  = window.innerHeight - headerSize               - tileMargin * (groupHeight - 1);
   var optimum = Math.min(width/groupWidth, height/groupHeight);
   return Math.min(maxTileSize, optimum).toFixed(1);
}
```

rchl:
@akloeckner it should still work. You should check the browser console for the exact error.

When I've tried this code here I got an error about undefined `suffix` so I guess you were relying on the global one from the old `index.html`?

akloeckner:
Ok. I replaced the `suffix`. Now, it's complaining about `angular` not being found. Also `window.angular` doesn't get found. I'm using `angular.equals` and `angular.element`...

BTW: Yes. I just prefer my setup to be tidy. So, I moved all helper functions outside of config.js. I just couldn't find my way around in that humongous file anymore...

rchl:
Right.

Due to this `document.write` nonsense, I'm creating two bundles, one with just essential global constants and utilities and another with the rest of the app. Config only has access to the former when it's evaluated and that former bundle doesn't have `angular` source yet.

Adding `angular` source to that first bundle would duplicate angular which means a much bigger bundle size and also some warnings from angular due to being evaluated twice.

I can't see an immediate clean solution for that but I can look more into it later if we are not fine with this being a breaking change.

akloeckner:
I could get rid of `element`. But I need this in a few places:
```
var id = window.angular.equals(id0, {}) ? undefined : id0;        
```

rchl:
> I could get rid of `element`. But I need this in a few places:
> 
> ```
> var id = window.angular.equals(id0, {}) ? undefined : id0;        
> ```

Such a simple check could probably easily be replaced with something like `Object.keys(id0).length === 0` but still this is something that should probably be addressed.

akloeckner:
I don't understand the process of bundling that you introduced. But would it help for this issue to have only one bundle? And what changes would be required on the user side? If the required changes are easily explainable, I'd say we do the restructure properly in the first place and put a migration section into the wiki.

rchl:
There are two bundles because the config needs to be loaded between those two. Those two bundles are:
 1. Constants and Global Utils
 2. The whole rest including angular, dependencies, and TileBoard.

Config needs to be loaded between those two as it needs access to constants like `TYPES.*` and all the other.
At the same time, the TileBoard needs access right away to `CONFIG` from `App.config` and such.

Making one bundle could work if Config would utilize `angular.service` and `angular.provider`. The config would be structured something like this:

```js
angular.module('App').provider('configProvider', function () { this.$get = function(Constants, Noty) { ... } })
```

I believe that should work. I haven't actually tried it yet.

Evgeny-:
👍

Have you tried to make app initializing asynchronous? I'm not sure what was the original problem, probably was too lazy to make it. It should help to those who want to make module configs. Though, without rebuilding the app. 

May be adding global `require` function could help solve that problem?

rchl:
Your comment reminded me that there is a way to [initialize angularjs manually](https://docs.angularjs.org/guide/bootstrap#manual-initialization). It actually works quite well with just a single bundle now. Thanks! :)

rchl:
I've reverted some unnecessary style changes to make diffs easier to review.
Style issues will be handled in a separate PR.

rchl:
In the latest commit, I have fixed issues with assets referenced from CSS not being included in the dist/ folder which meant that those wouldn't be available after deploying.

Also made production bundles have hashes in the filename. That means that on every change of some code, the hash will change which should eliminate issues with caching and not seeing the latest changes after updating to the new version. Also we don't have to bust cache on every request which allows the browser to cache those and load TB faster.

akloeckner:
I replaced the `angular.equals` by your suggestion. No issue anymore. Now it complains about me using `mergeObjects`, which is apparently found. But inside of it `window.angular.merge` is not found...

Evgeny-:
@resoai let's just merge it and solve all problems afterwards :)

resoai:
> @resoai let's just merge it and solve all problems afterwards :)

That is exactly our way! Since I still have absolutely no way to test it, I will leave it up to @rchl and @akloeckner to decide when it is time to merge.

rchl:
@akloeckner Actually with my latest changes all of `angular` should be available in your config. So you can keep using it.
As for `mergeObjects`, I've pushed small change for that but it should have worked already. Maybe try again and post exact error?

akloeckner:
Thanks! That seems to work now! (With original config, using `angular.merge` directly.)

When trying to start the development environment (impressive, I must say), I noticed a few thing to probably **add to the `CONTRIBUTING` docs**:
* `config.js` must be copied to `dist-dev` (not `dist`)
* `http://<your-dev-server>:8080` needs to be added to HA's `http: cors_allowed_origins:`
* in case you use https for HA, the default in `config.example.js` does not work anymore, because `dist-dev` is served on http

I have one open question still:

In my "production" environment I use multiple branches currently, which I merge together: `master` for sure, plus `secret` with my config, plus a few feature/fix branches for testing before I open PRs. As I understand, that will not be possible anymore, because merging does not execute the bundling step, right? (I guess, it boils down to the fact that I have local modifications of the code. An I'm sure, I'm not alone with this.) What would be the steps, I have to take? (Probably, I wouldn't want to override `dist`, because this would leave me with a dirty git repo. And I also wouldn't want to be running `npm run dev` all the time.) I suggest to **put those steps in the contributing docs** (or some more suitable place) as well. 

As to the huge amount of changes, I'll have to trust you @rchl , since I have no idea of the magic you have done!

So, apart from these documentation changes, I suggest to merge the PR. Looks promising!

akloeckner:
Another thought I just had: I have tileboard as a sub-repo in my HA config repo. So, instead of copying files from `dist`, I'll just open a different URL. The same might hold true for people simply checking out tileboard to their webserver. 

So, it might help to put a minimal index.html in the tileboard root explaining that the app has moved. It would point people with those kinds of setup in the right direction.

rchl:
Good foodback. Will try to address it later.

As for your development situation with different branches I'd have to understand that better. Would you be using `dist-dev` output or `dist`? With dist you would commit your modifications so changing branches shouldn't be an issue.

akloeckner:
More observations: 
* somehow the history graphs don't have tooltips anymore.
* some graphs fail with `r is undefined`
Screenshot_2020-09-26-17-46-40

* the endpoints of the graphs are strange.
IMG_20200926_174753

akloeckner:
As to different branches, i usually do something like this:
```
git checkout -b deploy master
git merge secret fix-camera-title cleanup-less
```

This leaves me with a working directory including all features I need/want. Will those merges work with `dist`? I have the impression, it will be a matter of luck, if merges will conflict, due to the toolchain inbetween. I don't mind changing my workflow. But I'd like it to be as easy (automated) as possible. I could extend my above script with some rollup command. Or I could set up an npm dev service.

Probably the branching thing is a specialty of mine. But others might have simple modifications/fixes they don't bother to put in a PR. Such as fixing the issue here #425 . 

So, suppose the user has fixed that one line from #425 locally, what would be the steps to bring this change into `dist`? And will it survive a `git pull`?

rchl:
With the latest changes the hash in the name of generated bundle changes whenever there is change to any script file. So that will not work with merges at all *. Unless git is smart enough to figure out that it's basically the same file with some small changes (which in fact it might be able to do).

I guess we could consider not committing `dist/` directory at all and instead automatically generating a release package on commit (through github actions). Then the end-user would find the zipped release on the [Releases](https://github.com/resoai/TileBoard/releases) page and developers would have to `npm run build` locally to generate their own version with modifications.

* I mean when it comes the `dist` directory of course. The changes in `/scripts` would still merge as well as before so you could still generate your release bundle easily from that.

rchl:
> some graphs fail with r is undefined

If you are able to reproduce it, can you check what response from API do you get when that happens?

akloeckner:
> Unless git is smart enough to figure out that it's basically the same file with some small changes (which in fact it might be able to do).
> 
> I guess we could consider not committing `dist/` directory at all and instead automatically generating a release package on commit

I agree. Git is probably smart enough to track changes through renames. But a merge will still fail, because the two *different* renames will create a conflict for sure...

Creating ZIPs automatically will probably make it even easier for more non-techy people to use Tileboard. 👍 

Possibly, it will make sense to also commit `dist`, but on a branch which is likely to not need merges, say `release`. That would allow for the average user to do a simple pull, instead of downloading ZIPs. But it would add quite some overhead to maintaining the repo. If I were you, I'm not sure, if I'd do that.

The developer user will probably be fine with running an additional command to build. I will be at least. (Actually, I'm looking forward to having the dev server, which will not break the production dashboard while I'm playing with new features for it.)

rchl:
> > some graphs fail with r is undefined
> 
> If you are able to reproduce it, can you check what response from API do you get when that happens?

I've managed to reproduce by restarting HA and then trying to run TB as early as possible after HA restarts.
This is possibly due to some timing issues.

It all goes south when due to this:

```js
               // Create extra state with current value.
               dataset.push({
                  x: Date.now(),
                  y: $scope.states[firstStateInfo.entity_id].state
               });
```

That last state seems to be a bit broken in this case:

```
x: 1601148398645
y: "unknown"
```

So it looks like `$scope.states` doesn't have up to date information at that point. Not sure it's a new problem...

rchl:
@akloeckner 

> somehow the history graphs don't have tooltips anymore.
> some graphs fail with r is undefined
> the endpoints of the graphs are strange.

Turns out `angular-chart` was bundled with old version of `chart.js`. Fixed all those issues by forcing the latest version of `chart.js`.

Note that even there is no crash anymore, the #434 is still valid to fix the issue with a wrong graph display.

rchl:
> Another thought I just had: I have tileboard as a sub-repo in my HA config repo. So, instead of copying files from `dist`, I'll just open a different URL. The same might hold true for people simply checking out tileboard to their webserver.

I've added an `index.html` in the root.

> * `config.js` must be copied to `dist-dev` (not `dist`)

Changed

> * `http://<your-dev-server>:8080` needs to be added to HA's `http: cors_allowed_origins:`

Added a note about that. Unfortunately, that won't work for history API right now (https://github.com/home-assistant/core/issues/39727).

> * in case you use https for HA, the default in `config.example.js` does not work anymore, because `dist-dev` is served on http

Added a note on setting that.

akloeckner:
> Created #434 for it.

My HA is runing for days. So this is probably not my very issue. (Still good, it's identified!) My debugging pointed me to some tooltip code. (`tooltip` is replaced by `r` in the bundled code, but the dev server still had `tooltip`.) So, I hope, this problem will be solved by you having forced the latest Chart.js version.

> Fixed all those issues by forcing the latest version of chart.js.

Could you commit that change , so I can test on my end again?

> Unfortunately, that won't work for history API right now (home-assistant/core#39727).

Well, actually, for me that *does* work, as I noted in #386. However, good it's in the docs now. I'd still suggest to also give a hint to the HA setting, because that's what I had to do to be able to authenticate with HA, even.

akloeckner:
> Could you commit that change , so I can test on my end again?

Sorry... I didn't see that commit. Tested and all seems to be fine now! Thanks!

rchl:
> I'd still suggest to also give a hint to the HA setting, because that's what I had to do to be able to authenticate with HA, even.

Thanks for the remainder. Forgot about it. Now it's added.

> EDIT: Tested in dev server. Very neat. But had to rebuild the `dist` using `npm run build`.

Yes, every change in the scripts or dependencies requires rebuild.

I think this should be solved by forcing build on every commit but since we've decided to investigate not committing the `dist` folder, I'll look into that instead.

akloeckner:
> CORS...

Thanks! Sorry for being pedantic: `localhost` will not always be the right host. Probably developers will notice. But since we're discussing it here... I'd phrase it:

Configure `http.cors_allowed_origins` setting in Home Assistant to allow your server (e.g. localhost) to communicate with the HA API:

```yaml
http:
  cors_allowed_origins:
    - http://<your-server>:8080
```

> every change in the scripts or dependencies requires rebuild.

Which takes 2m18s on my pi! 30s for the dev server. Maybe we can speed this up once we have the system in place. In my opinion that's anything but "iterate quickly"... Also, it seems to write lots of files. Not sure, what it will do to my SD card wear...

rchl:
> Which takes 2m18s on my pi! 30s for the dev server. Maybe we can speed this up once we have the system in place. In my opinion that's anything but "iterate quickly"... Also, it seems to write lots of files. Not sure, what it will do to my SD card wear...

Personally, I develop on my main Desktop and have a script that just pushes built files to PI.

You could speed up production build by removing `terser` plugin (`appPlugins.push(terser());` line) which is for minifying the code. This is probably what makes the biggest difference. Or maybe try with `treeshake: false,` in the config.

But yeah, it can be a bit heavy. There are pros and cons to everything...

akloeckner:
> Personally, I develop on my main Desktop

I should start that, too, maybe... :-) I'll try your options and see, what it does.

akloeckner:
I now also got around to testing this with my old iPad, iOS 9.35. I had to create a long-lived token for it (see https://github.com/home-assistant/frontend/issues/5376). However, the TileBoard does not load. It is stuck on the background color. Usually that is an indicator for some broken JS code, hindering the app to initialize.

TileBoard works as usual in the old structure.

rchl:
Is it possible to get access to the console and see what it errors on exactly?
It's probably possible to configure builds process to support it.

akloeckner:
I'm not sure, how to access the console. The internet seems to suggest that I need Safari on my Laptop and some USB cable. It might take some time to get that up and running...

Looking at this issue, however, the `const` seems to be unsupported by the old iOS (at least):
https://github.com/rollup/plugins/issues/190

akloeckner:
> I'm not sure, how to access the console.

Hmm... as it seems, I need Safari for this to work. 
https://www.kenst.com/2019/03/how-to-debug-problems-on-mobile-safari/

Unfortunately, Safari is said to be only available on Mac.
https://de.wikipedia.org/wiki/Safari_%28Browser%29

But I have no Mac. So, I guess I'm stuck...

rchl:
I Will take care of the `const` problem and we'll see. The code is supposed to be transpiled for old browsers so there shouldn't be any `consts` in the production bundle

rchl:
OK, pushed a change that actually transpiles the code for older browsers. Not sure if it will handle your old iOS by default. If not then we'll have to change config to target older browsers.

rchl:
OK, some bigger changes in the latest commits:
 - `dist` directory is not committed (no built app is included)
 - both `npm run build` (production build) and `npm run dev` output to `build` directory (so not `dist` and `dist-dev` respectively)
 - New release will be created using `npm version patch` and that will trigger a github workflow which will automatically build, package and create a new release on https://github.com/resoai/TileBoard/releases. Then people can just download the zip file from there that includes only built code and nothing else.

akloeckner:
I checked my old iOS. It works (`npm run build`)!

Since everything works now on my end, I'll be fine to merge as soon as you (and @resoai) are happy with the workflows!

rchl:
I'm happy with the changes myself too so will merge later in the day.

PR #435, opened 2020-09-28
Fix linting issues and enable more eslint rules to enforce consistency

rchl:
Those changes were done mostly automatically by `eslint --fix` command.
There should be no functional changes.

@akloeckner


resoai:
LGTM

PR #449, opened 2020-10-01
fix: camera title extending beyong tile boundaries

### LINKED ISSUE #429: Title outgrowing tile for camera
I just noticed that the title overflow is not hidden for camera tiles (and a few more). Is there a specific reason for it? 

https://github.com/resoai/TileBoard/blob/538d00a40a7418f4058cd3eae795e676a0ade9e6/styles/main.css#L618

akloeckner:
fixes #429

### DIFF
--- a/styles/main.less
+++ b/styles/main.less
@@ -649,6 +649,7 @@ body {
          text-shadow: 1px 1px 0 rgba(0,0,0,.2);
          background: rgba(0,0,0,0.15);
          right: auto;
+         max-width: calc(100% - 6px - 12px);
          line-height: 22px;
          height: 22px;
          left: 3px;


styles/main.less
```diff
@@ -649,6 +649,7 @@ body {
          text-shadow: 1px 1px 0 rgba(0,0,0,.2);
          background: rgba(0,0,0,0.15);
          right: auto;
+         max-width: calc(100% - 6px - 12px);
```
rchl, line 652: I think you can instead just set `right: 3px`.

If that works then that's preferred since `calc` is a fairly new feature so should be used sparingly in this project that aims to be compatible with some old crap devices.

rchl: No, sorry, that would not be equivalent when text is shorter than width.

Would be easy to fix with plain css by adding an extra wrapper element but I won't require that here.

rchl (approved):

PR #450, opened 2020-10-01
refactor: use modular popup for iframe popups

akloeckner:
This is another refactor to make maximum use of the modular popup with the following changes
* use modular popup for iframe, too
* add classes to make it easier to specify size of popups and tiles inside them
* remove old code from stand-alone iframe popup

BREAKING CHANGE: This changes some behavior:
1. background of iframe tile is changed to ~~transparent~~ white, because otherwise, the background of the iframe would be awkwardly gray.
2. ~~`iframeClasses` are applied to `popup-container`. Not `popup` (including the blurred-out area around the popup). This seemed more intuitive. But I think, it could be changed to less intuitive `popup` to not break existing setups. (And possibly allow for styling the blurred-out area.)~~
3. `iframeStyles` are applied to the iframe tile. Not to the `popup-container`. This is mainly due to the way, the popup is structured. There simply *is* a facility to pass styles to a tile. But there is *not* a facility to apply styles directly to the popup. For the user, this means that styles (e.g. borders) do not include the popup title anymore. 

Fixes to these changes:
1. set background manually if need to.
2. ~~fix `custom.css` styles to reflect new structure~~
3. use `iframeClasses` together with `custom.css` and forget about `iframeStyles`

In summary, I would favor these changes over implementing the old structure and believe that they are not of a very big impact to the user. Especially the `styles` options should (in my opinion) be disfavoured in comparison to using `classes` options. This will make it possible for us to supply even more classes for easy customization.


rchl:
> 1. background of iframe tile is changed to transparent, because otherwise, the background of the iframe would be awkwardly gray.

Can we set background to white by default? Transparent is possibly not the best choice as then many pages (those that don't define background color) will be transparent. Browsers use white background by default so matching that might be good.

scripts/controllers/main.js
```diff
@@ -1560,10 +1527,24 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       };
    };
 
+   function initPopupLayout (item, entity, key, layout) {
+      if (!item[key]) {
+         item[key] = layout;
+      }
+      return $scope.openPopup(item, entity, item[key]);
+   }
```
rchl, line 1535: I wouldn't expect the **init**PopupLayout to actually open the popup, I think.

But also, can we just have this `if (!key) item = layout` logic within the `openPopup*` functions themselves? I don't think we are gaining much by having common functions like that and also we are being less efficient because the layout still has to be created every time before this function is called, even if the initial layout is gonna be reused here.

akloeckner:
> Can we set background to white by default?

Hmm... I am wondering now, why that is not an issue with my page that became gray. I'll investigate and switch to white in the process.

rchl (commented):
> > Can we set background to white by default?
> 
> Hmm... I am wondering now, why that is not an issue with my page that became gray. I'll investigate and switch to white in the process.

Because there are various aspects to it. It depends on the theme and also on whether we are talking about the iframe tile or iframe popup.

Here is tile:
Screenshot 2020-10-01 at 20 33 47

Here is popup:
Screenshot 2020-10-01 at 20 33 57

Neither is really great as the tile became transparent and the popup background is black (or close) so the text is unreadable.
This is tested on http://elg.no page

styles/main.less
```diff
@@ -1374,6 +1359,42 @@ camera_stream {
...
 
+// the following classes can be used to style the popup-container
+.-popup-fullsize {
+   .popup--tiles {
+      /* use classes: ['-popup-fullsize'] to fullsize the popup */
+      width: 100vw;
+      height: 100vh;
+      max-width: calc(100% - 2*@popupPadding);
+   }
+}
```
rchl, line 1370: Since you are in `less` and have possibility to nest things, this is how it should be done:
```diff
--- a/styles/main.less
+++ b/styles/main.less
@@ -1356,18 +1356,17 @@ camera_stream {
       width: 100vw;
       /* debug:
       border: 0px solid red; */
+
+      .-popup-fullsize & {
+         /* use classes: ['-popup-fullsize'] to fullsize the popup */
+         width: 100vw;
+         height: 100vh;
+         max-width: calc(100% - 2*@popupPadding);
+      }
    }
 }
 
 // the following classes can be used to style the popup-container
```

This change is in this block: https://github.com/akloeckner/TileBoard/blob/250a2d74ed092550a76e83c46fddbe539a86954d/styles/main.less#L1327-L1360

This way you keep all related classes together and make it easier to see what overrides what.

scripts/controllers/main.js
```diff
@@ -1742,30 +1723,51 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-               width: '100vw',
-               height: '56vw',
-               margin: 0,
-               maxWidth: '100%',
-            },
+      return initPopupLayout(item, entity, '_popupHistory',
+         {
+            classes: ['-popup-landscape'].concat(
+               getItemFieldValue('history.classes', item, entity) || [],
+            ),
```
rchl, line 1730: We can use some modern JS features now like destructuring:
```suggestion
            classes: ['-popup-landscape', ...(getItemFieldValue('history.classes', item, entity) || [])],
```
Shorter and cleaner.

akloeckner: nice trick. done

styles/main.less
akloeckner: another nice trick. :-) also done

akloeckner:
> Can we set background to white by default? 

Did so. I was offset by the fact that my test iframe had an `iframeStyles` option, styling things white already...

scripts/controllers/main.js
akloeckner: I have seen quite some slightly different implementations of the same idea, when it comes to this type of caching. I thought, it would be wise to have it done once. And have it done right. So, new popups could re-use. When investigating the issue #425, I came up with a solution that would allow to also avoid expensive calculations (like `getHistoryObject`). So, by now, I would suggest something more generic:
```
   function cacheInItem (item, key, data) {
      if (!item[key]) {
         item[key] = typeof data === 'function' ? data() : data;
      }
      return item[key];
   }
```

We could use it as follows:
```
return $scope.openPopup(item, entity, cacheInItem(item, key, {
   // layout goes here
}));
```

Obviously, we would not spare generating the layout objects multiple times. (Unless some shiny new features allows something like this:
```
return $scope.openPopup(item, entity, cacheInItem(item, key, () => {
   // layout goes here
}));
```

What do you think?

akloeckner:
I also styled the whole popup. It will allow to define, e.g., `alert` classes for popups, that style the whole thing red, including the blurred area. It's now consistent with the old implementation.

One thing that came back to my mind now is, that `history.classes` is now actually used for two elements:
1. the `popup` as introduced here. (This is actually also in the docs, but I forgot about that when migrating the history popup.)
2. the `tile` as the `history` object is merged into the generated `tile` object.

It should not be a big deal, I believe, because the user can always implement `custom.css` in a way that their classes need to be applied to a `.popup` or a `.item`. But it clearly might be a bit confusing.

scripts/controllers/main.js
rchl: Let's pass the function. That will spare the generation if it's already cached.

For readability, I would suggest not cramming everything on the same line though but do something like:

```js
const layout = cacheInItem(item, key, () => {
   // layout
})
return $scope.openPopup(item, entity, layout)
```

BTW. Can we always use the same cache key instead of using different ones like `_popupHistory` and `_popupIframe`. Can we just use `__layout` always? Then the `initPopupLayout` function will have slightly more reason to be.

scripts/controllers/main.js
```diff
@@ -1742,30 +1723,47 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-         item[key] = {
-            styles: {
-               width: '100vw',
-               height: '56vw',
-               margin: 0,
-               maxWidth: '100%',
-            },
+      return initPopupLayout(item, entity, '_popupHistory',
+         {
+            classes: ['-popup-landscape', ...(getItemFieldValue('history.classes', item, entity) || [])],
```
rchl, line 1728: BTW. This is spread, not destructuring, as I've erroneously said before. :)

styles/main.less
```diff
@@ -1371,6 +1356,39 @@ camera_stream {
...
+   right: -@popupPadding;
+   z-index: 0;
+
+   &:active {
+      margin: -@popupPadding !important;
+      border: none !important;
+   }
+
+   .item-clickable {
+      display: none ! important;
```
rchl, line 1391: Should using `.-tile-fullsize` imply that the item is not clickable. I can imagine someone wanting a popup with a single, fullsize button.

rchl:
Heads up: I might not be able to finish reviewing it this weekend as I'll be away.

scripts/controllers/main.js
akloeckner: * passing function: alright
* multiple lines: alright
* `initPopupLayout`: I think, I would get rid of it then. The purpose was to provide a proper and standard way to cache and open the popup. We don't need that anymore with `cacheInItem`. I would trust a developer to copy the simple structure to a new popup. And it does not save us (m)any lines of code...
* unique cache key: I'm not sure. At least history will have a problem with that, because it defines a watcher modifying the data stored in the `_popupHistory....historyTile`. It will be difficult to catch that properly, I believe. At least, as long as we expect tiles to be able to open more than one popup. Which we should provide for, since we have primary and secondary action.

scripts/controllers/main.js
akloeckner: :-)

styles/main.less
akloeckner: At least, it looks weird. We could attach a class `.-popup-history` to the popup and specialize the code for that?

akloeckner:
> Heads up: I might not be able to finish reviewing it this weekend as I'll be away.

Fine. That means more time for my family. :-)

styles/main.less
rchl: Maybe add a modifier class like `-non-clickable` and use it for the popup history to make the intentions clear.

akloeckner: good idea. done

scripts/controllers/main.js
akloeckner: adressed the first three points. 

still not sure as to the fourth. I suggest to keep separate keys, because I am not sure of the side effects of deleting a popup layout with its caches.

scripts/controllers/main.js
```diff
@@ -1564,6 +1531,13 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       $scope.activePopup = null;
    };
 
+   $scope.getPopupClasses = function () {
+      if (!$scope.activePopup || !$scope.activePopup.layout.classes) {
+         return [];
```
rchl, line 1536: Would returning `null` here work fine?

While it's probably not a big deal either way, returning a new array instance is likely more work for angularjs than returning  `null` as in former case it will see the value as being changed and trigger watchers (also internal ones) while with `null` it should not need to do that.

That's also probably why the original code has used a single object instance and just removed and added properties on it.

styles/main.less
```diff
@@ -1371,6 +1356,40 @@ camera_stream {
...
+      .-popup-landscape & {
+         /* use classes: ['-popup-landscape'] for a landscape oriented full-width popup */
+         /* be sure to also set styles: {}, in order to avoid size calculations */
+         width: 100vw;
+         height: 56vw;
+      }
+   }
+}
+
+.-item-fullsize {
```
rchl, line 1376: This is specific to tiles within the popup so should IMO be moved inside `.popup--tiles` selector.

styles/main.less
```diff
@@ -1371,6 +1356,40 @@ camera_stream {
...
+   z-index: 0;
+}
+.-item-non-clickable {
+   /* use classes: ['-item-non-clickable'] to display as non-interactive tile */
+   &:active {
+      margin: 0 !important;
+      border: none !important;
+   }
+   .item-clickable {
+      display: none ! important;
```
rchl, line 1392: This is currently unused, right? What would be the use case for it?

Also, the use of `!important` is generally a bad style and often source of issues later. Can we just make the selector more specific to be able to avoid using `!important`? And can we move it close to original styles that it is supposed to override?

rchl (changes requested):

styles/main.less
akloeckner: Just to be sure: what would be good style for less generally? I personally like to have things together that belong to each other. Also I would prefer not too much nesting. This seems to go against the general less idea. 

I am e.g. very much confused by specific tile styles (and also general pop up styles btw) spread over several places. Currently there are some parts of it (clickable, off-states) within the general item selector. And several parts are elsewhere. I would prefer to have one block to define the general concept of a tile/item. Possibly with general modifier classes. And to have one separate block for each specific tile type. Possibly extending the general modifier classes. 

Following these ideas, here, we have just a single modifier class, which probably only has a meaning within a popup. So, probably I would agree to add it inside popup—tiles. But I’d like to understand if you agree with the reasoning? (Which seems a bit different from yours.)

styles/main.less
akloeckner: It’s unused and I will remove it here. Just to introduce it again for the doorentry popup. ;-) Maybe then as a child of `.popup—tiles .item`.

styles/main.less
rchl: > I personally like to have things together that belong to each other. Also I would prefer not too much nesting. This seems to go against the general less idea.

I'd agree that avoiding nesting is against LESS. In my view the main advantage of LESS (or SASS/SCSS) over CSS is that you can nest things which make it easier to see the structure and know which styles apply where. With CSS that would only sort of work when you follow very strict naming rules (like BEM).

I think the current code is a bit of a mess so I wouldn't really try to follow existing patterns. For example, it's sort of a BEM style but not really consistent.

Any code refactoring now has a chance of breaking people's overrides so I wouldn't be too eager in changing existing code but new code can follow some common logic.

scripts/controllers/main.js
akloeckner: I just checked the docs. It says nothing about null.

https://docs.angularjs.org/api/ng/directive/ngClass

I also did quite some reading on what angularjs actually watches. And if I am not mistaken, our function is re-evaluated often anyways. Plus, using a function makes angular apply a deep watch. So, a “buffer” array will not improve the speed. See the last comment here:

https://github.com/angular/angular.js/issues/15905

We could return a comma-separated string, which should be cheap to watch. What do you think?

rchl: > And if I am not mistaken, our function is re-evaluated often anyways.

Yes, the function is gonna be evaluated regardless but if the result of the function doesn't change then angularjs would not have to notify the watchers. And it does change every time on returning an empty array. Returning `null` would allow angularjs to see that nothing has changed since null always equals null. But since ng-class doesn't officially support `null` then I guess we should avoid that.

akloeckner: You marked this resolved. Don't you think, I should return the empty string `''` here? That should always compare to the last empty string, avoiding any watcher executions, right?

rchl: Yes, that should be good too. :)

rchl:
Thanks!

PR #461, opened 2020-10-08
refactor: use modular popup for door entry

### LINKED ISSUE #447: Camera doesn't catch fullscreen undefined
The `fullscreen` option is not defined as mandatory, but it generates an error, if it not defined and we click on the tile. Found that error, because I was wondering why camera tile is "clickable".

That line should be change to use `getItemField`:
https://github.com/resoai/TileBoard/blob/c2898220f0f09ee15950801026d2d1727fe7d625/scripts/controllers/main.js#L171

akloeckner:
Here comes the next refactor to make more use of the modular popup.
* use modular popup and remove legacy code
* make camera not freeze when on popup
* add timeout to modular popup

I don't see any breaking changes with this.

Fixes #447

### DIFF
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -83,43 +83,6 @@
       </div>
    </div>
 
-   <div class="door-entry-popup"
-        ng-if="activeDoorEntry && (page = activeDoorEntry.layout.page)
-            && (cameraEntity = getEntryCameraEntity(activeDoorEntry))">
-
-      <div class="door-entry-popup-container">
-         <div class="door-entry-popup-title">
-            <div class="door-entry-popup-close" ng-click="closeDoorEntry()">
-               <i class="mdi mdi-close"></i>
-            </div>
-            {{ entityTitle(activeDoorEntry, entity) }}
-         </div>
-
-         <div class="door-entry-popup-container">
-            <div class="door-entry-popup--camera">
-               <div ng-if="activeDoorEntry.layout.camera.type === TYPES.CAMERA">
-                  <camera item="activeDoorEntry.layout.camera" entity="cameraEntity"
-                          freezed="false"></camera>
-               </div>
-
-               <div ng-if="activeDoorEntry.layout.camera.type === TYPES.CAMERA_THUMBNAIL">
-                  <camera-thumbnail item="activeDoorEntry.layout.camera" entity="cameraEntity"
-                                    freezed="false"></camera-thumbnail>
-               </div>
-
-               <div ng-if="activeDoorEntry.layout.camera.type === TYPES.CAMERA_STREAM">
-                  <camera-stream item="activeDoorEntry.layout.camera" entity="cameraEntity"
-                                    freezed="false"></camera-stream>
-               </div>
-            </div>
-
-            <div class="door-entry-popup--tiles">
-               <div tile ng-repeat="item in activeDoorEntry.layout.tiles track by $index"></div>
-            </div>
-         </div>
-      </div>
-   </div>
-
    <div class="popup"
         ng-if="activePopup && (page = activePopup.layout || {})"
         ng-class="getPopupClasses()"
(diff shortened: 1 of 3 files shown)


rchl:
Interestingly, I've created a tile to test this and it works with these changes but not on master. Do you have any idea why? I don't get any errors on master, just clicking the tile doesn't open the popup.

```js
                  {
                     position: [6, 6],
                     width: 2, height: 2,
                     type: TYPES.DOOR_ENTRY,
                     id: {},
                     icon: 'mdi-phone',
                     title: 'Door entry',
                     state: false,
                     layout: {
                        camera: {
                           type: TYPES.CAMERA,
                           objFit: 'contain',  // https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
                           id: 'camera.front_gate',
                           refresh: 1500,
                           bgSize: 'cover'
                        },
                        page: {},
                        tiles: [
                           {
                              position: [0, 0],
                              // width: 2, height: 2,
                              theme: ITEM_TRANSPARENT,
                              title: 'SCRIPT',
                              id: 'script.test_script',
                              type: TYPES.SCRIPT,
                              state: false,
                              variables: {
                                 value: 'value from script',
                              },
                           },
                        ]
                      }
                  },
```

rchl:
Actually, I've checked myself and it's because such a camera ID doesn't exist in my states (returned from `getEntryCameraEntity`). I guess previous behavior was worse as it didn't give any indication of what's wrong.

rchl:
My test tile looks a bit different due to the tile's theme being applied in new code:

OLD:
Screenshot 2020-10-08 at 23 23 29

NEW:
Screenshot 2020-10-08 at 23 23 57

The camera I'm using is a test camera taken from documentation:

```yaml
camera:
  - platform: generic
    name: Weather
    still_image_url: https://www.yr.no/place/Norway/Oslo/Oslo/Oslo/meteogram.svg
    content_type: 'image/svg+xml'
```

rchl:
Clicking on the camera triggers:
```
TypeError: Cannot read property 'id' of undefined
    at Scope.$scope.getCameraEntityFullscreen (main.js:168)
```

akloeckner:
* **popup now opening with invalid camera**: We could catch it in `openDoorEntry`. I'd call the current behavior a bug, though. It even leaves the internal `activeDoorEntry` variable in a messy state.
* **tile background**: That's ugly. I should change it. Will setting a transparent background be sufficient? I could add that to the `item-fullscreen`, e.g.
* **`fullscreen` undefined**: That's a general bug of the camera tile. See #447. But obviously, it's most prominent here. We could quick'n'dirty override `action` in the main popup tile. 
* **open second popup**: That could have happened before already, if you put a camera tile in the `tiles` of that popup. But yes. That should be adressed. I suggest to generally close all (fullsize/modal) popups before opening a new one. And possibly re-arrange the code for the remaining popups (e.g. datetime) such that those are on top of fullscreen popups.

What do you think?

rchl:
I've accidentally merged before checking it out as I've used `gh` cli tool that I've expected to merge to local branch and not to master. I will check it after the fact then and report if I see any issues with it. :)

akloeckner:
:-) I was surprised, already.

rchl:
This tile doesn't open fullscreen popup on clicking:

```js
                  {
                     position: [6, 6],
                     width: 2, height: 2,
                     type: TYPES.CAMERA_THUMBNAIL,
                     // theme: ITEM_TRANSPARENT,
                     objFit: 'contain',  // https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
                     id: 'camera.weather',
                     refresh: 25000,
                     bgSize: 'contain',
                     fullscreen: {
                        type: TYPES.CAMERA,
                        refresh: 15000, // can be number in milliseconds
                        bgSize: 'contain'
                     },
                  },
```

akloeckner:
I pushed the fix to my branch. But not sure if it will show up in the PR still...

PR #468, opened 2020-10-12
fix: bundle with mdi fonts instead of loading from external server

### LINKED ISSUE #453: Suggestion: Offline MDI
MD icons are currently loaded from the web. This might cause some issues, like #426 I believe. Also, this is the only resource not bundled with Tileboard. I suggest to include the icons in the bundle, if possible.

It seems, this might be a suitable approach: https://google.github.io/material-design-icons/#icon-font-for-the-web (second method)

rchl:
Resolves #453

### DIFF
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -1090,10 +1090,5 @@
    </div>
 </script>
 
-
-
-<link rel="stylesheet" async
-      href="https://cdn.materialdesignicons.com/5.3.45/css/materialdesignicons.min.css">
-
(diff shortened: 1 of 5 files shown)


rchl:
@akloeckner

akloeckner:
Nice! :-) I have no clue as to the correctness of that code, however...

rchl:
Can be proof-checked by just looking at the generated output in `build/styles/styles.css`. :)

PR #470, opened 2020-10-12
fix: cache in item instead of entity

### LINKED ISSUE #425: Scripts with same id overwrite background

### LINKED ISSUE #464: TYPES.SLIDER and 0 values.
When using a slider tile. Values with 0 show visually as the slider being in the middle of the tile (50%). This doesn't accurately represent the value. For example to set to 10% I have to move the slider backwards.

I have been updating the code as follows to fix this. It would be good to somehow integrate this into the code. Either for everyone, or as perhaps with a boolean value on the tile to tell it how to display 0

getSliderConf 

         //var value = +attrs[def.field] || 0;
         var value = +attrs[def.field] || +entity.state || 0;

         //value: value || +entity.state || def.value || 0, 
         value: (value === 0) ? (0) : (value || +entity.state || def.value || 0),

akloeckner:
fixes #464 
fixes #425 

@rchl , I have changed `eslint` to accept the `??` operator as of ECMA Script 2020. It transpiles just fine. But I am not aware of the issues this might cause on the rest of the linting.

Explanation of that operator:
https://github.com/tc39/proposal-nullish-coalescing

PS: Regarding the breaking change below, we should take a short time and think about which order is more suitable `def ?? attrs` or  `attrs ?? def`. I tend to think now that `def` should have priority, because it is user-defined. 

BREAKING CHANGE: The order of `attrs` and `def` is changed for a few sliders. This should not normally be a problem, because the user would only use one of the possibilities. But... just to give note on that.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,6 +1,6 @@
 env:
   browser: true
-  es2017: true
+  es2020: true
   node: true
 parserOptions:
   sourceType: module
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -562,7 +562,7 @@
          <div ng-if="item.controlsEnabled" class="item-entity-sliders"
               ng-click="preventClick($event)">
             <div ng-repeat="slider in item.sliders track by $index"
-                 ng-if="(_c = getLightSliderConf(slider, entity))"
+                 ng-if="(_c = getLightSliderConf(item, entity, slider))"
                  class="item-slider-container">
 
                <div class="item-slider-title" ng-if="slider.title">
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -449,35 +449,22 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       return item.styles;
    };
 
-   $scope.itemBgStyles = function (item, entity) {
-      const obj = entity.attributes || entity;
-
-      if (!obj.bgStyles) {
-         let bg;
-         const styles = {};
-
-         if ('bgOpacity' in item) {
-            styles.opacity = parseFieldValue(item.bgOpacity, item, entity);
-         }
-
-         if (item.bg) {
-            bg = parseFieldValue(item.bg, item, entity);
-
-            if (bg) {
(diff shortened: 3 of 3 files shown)


rchl:
> @rchl , I have changed `eslint` to accept the `??` operator as of ECMA Script 2020. It transpiles just fine. But I am not aware of the issues this might cause on the rest of the linting.

That's fine. It only affects linting.

Did you mean to include all those changes in this PR?

akloeckner:
> Did you mean to include all those changes in this PR?

It was kind of the same "issue" for me. As you can see, I did the first version of most of the changes in the very first commit...

We could split in three, but it will be some work:
* fix the background to use `cacheInItem`
* consolidate the sliders to use it as well
* fix the zero-value-error.

Do you prefer that?

rchl:
So...

I would be fine with multiple semi-related commits but then I'd avoid squashing all commits together on merging so that they are listed separately in changelog and history. But in this case, those separate commits are not really arranged or formatted properly as you've added something in one commit to remove or change it later. And some commits don't have a summary that follows conventional commit convention.

Basically, when looking at individual commits from oldest to latest, each should be a self-contained change that can work on its own.

So if you think cleaning that up is too much work, then I'd rather squash the changes but I wouldn't want to squash those two separate fixes together so I'd ask you to at least separate it into two commits, one for changing the `itemBgStyles` stuff and one for fixing the sliders stuff.

PR #471, opened 2020-10-12
fix: don't cache icon background

### LINKED ISSUE #425: Scripts with same id overwrite background

akloeckner:
fixes #425

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -440,13 +440,16 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.itemBgStyles = function (item, entity) {
-      const bg = getItemFieldValue('bg', item, entity);
-      const bgSuffix = getItemFieldValue('bgSuffix', item, entity);
+      return cacheInItem(item, 'bgStyles', () => {
+         const bg = getItemFieldValue('bg', item, entity);
+         const bgSuffix = getItemFieldValue('bgSuffix', item, entity);
+         const opacity = getItemFieldValue('bgOpacity', item, entity);
 
-      return cacheInItem(item, 'bgStyles', {
-         opacity: getItemFieldValue('bgOpacity', item, entity) || null,
-         backgroundImage: bg ? 'url(' + bg + ')'
-            : bgSuffix ? 'url(' + toAbsoluteServerURL(bgSuffix) + ')' : null,
+         return {
+            opacity: opacity || null,
+            backgroundImage: bg ? 'url(' + bg + ')'
+               : bgSuffix ? 'url(' + toAbsoluteServerURL(bgSuffix) + ')' : null,
+         };
       });
    };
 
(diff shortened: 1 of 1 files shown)


scripts/controllers/main.js
```diff
@@ -440,34 +440,17 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-      return obj.bgStyles;
+      return cacheInItem(item, 'bgStyles', () => {
+         const bg = getItemFieldValue('bg', item, entity);
+         const bgSuffix = getItemFieldValue('bgSuffix', item, entity);
+         const opacity = getItemFieldValue('bgOpacity', item, entity);
+
+         return {
+            opacity: opacity || null,
+            backgroundImage: bg ? 'url(' + bg + ')'
+               : bgSuffix ? 'url(' + toAbsoluteServerURL(bgSuffix) + ')' : null,
```
rchl, line 451: It's safer to keep the double quote. Otherwise, a `(` in the path (for example) would break parsing.

While at it, you can also switch to template literal:
```
`url("${toAbsoluteServerURL(bgSuffix)}")`
```

rchl:
Actually, I've just realized some potential issues with it. The background of the entity can change dynamically - for example, media player tiles' background should change on the next song. I think this will break that.

scripts/controllers/main.js
akloeckner: done

akloeckner:
That's true. But it was broken before then already. Nice thing about the new implementation with `cacheInItem` is that we can simply remove that function call and be good... ;-) 

Shall we? I don't know the inner workings of AngularJS well enough. 

We could also add a dummy cache, which is updated constantly, but will prevent any watchers from executing. Or do we need the watchers to execute in order to update the background? Seems a bit overdone to me, anyways.

rchl:
It wasn't broken before. Probably because the `entity` was recreated every time there was an update so caching didn't really do much in practice.

If we return a new object every time this function is called then angular will notify all watchers, even the values didn't really change. And that means that the DOM style property will be updated every time. I guess that was happening before anyway and the browser will actually optimize it so that nothing really happens. So while ideally notifying listeners would be preferred, I'm fine with just returning a new object everytime.

PR #473, opened 2020-10-12
fix: enable arm buttons based on supported alarm features

### LINKED ISSUE #365: Customize alarm action selection
My alarm system does not have an "Alarm Night" style behavior (to the best of my knowledge). Would be good to choose which actions are available to support various configurations.

rchl:
Resolves #365

### DIFF
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -147,22 +147,22 @@
 
             <div class="alarm-popup-buttons-line">
                <div class="alarm-popup-button -icon -home"
-                    ng-if="supportsFeature(FEATURES.ALARM.ARM_HOME, alarmEntity) && alarmEntity.state !== 'disarmed'"
+                    ng-if="supportsFeature(FEATURES.ALARM.ARM_HOME, alarmEntity) && alarmEntity.state === 'disarmed'"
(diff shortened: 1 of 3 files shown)


rchl:
@akloeckner Care to review?

akloeckner:
Sure! I'll be able to test earliest Thursday, though...

akloeckner (approved):
Actually, my `manual` alarm panel supports all those features. So, not much to test in the end. Looks good! 

If we were to clean up while passing by, I'd fix up the `isDisarmed` function to simply
```
return entity.state != 'disarmed';
```

rchl:
Removed `isDisarmed`. It doesn't add much and I would prefer to have as few scope methods as possible.

akloeckner:
:+1: My parameter for a good PR also is "more lines removed than added". ;-)

Just a logical twist, that strikes my mind: Is it really correct to enable the `arm` actions only, if the alarm is already `armed`. (And vice versa.) It seems to work correctly, but I don't understand, why...

rchl:
No, you are right. It's backwards and doesn't work as intended. Will fix.

akloeckner (approved):
I hope you weren't waiting for my approval...

rchl:
Was planning to merge eventually but was not in a rush. :)

Issue #312, opened 2019-12-01
behavior of long press for lights? 


akloeckner:
What is the intended behavior for long-press of (at least) lights? I experience that on long-press my light switches on. And a bit unreliably, the light slider is shown. I'd expect the light slider to show without the light switching on.

rchl:
I've also noticed that showing brightness slider on long-press is not very reliable. Investigate it a bit in the past and noticed that handling is a bit hacky since it relies on hardcoded timeout.

> I'd expect the light slider to show without the light switching on.

Devices don't report brightness (and other things like color) when off. At least both the Zigbee and Zwave devices that I have don't. You can see that even in Homeassistant those controls disappear when light is turned off.

akloeckner:
> handling is a bit hacky 

Maybe comparing it to the history popup can shed some light. I am not having any unintended clicks from this one.

> Devices don't report brightness (and other things like color) when off. At least both the Zigbee and Zwave devices that I have don't. You can see that even in Homeassistant those controls disappear...

While the former is true, the latter is not. I'm seeing the slider also when the light is off. See below. Which is most useful, as I can switch it on with a lower brightness directly. Tileboards slider also works perfectly like this. If it weren't for the ghost clicks...

![IMG_20191202_083623](https://user-images.githubusercontent.com/7125061/69938948-e7a70780-14de-11ea-9144-200df388c1c4.png)

rchl:
Is it Zigbee, Zwave or something else?

akloeckner:
It's Homematic. But it has no brightness attribute when off, either...

akloeckner:
I just found that the light is even turned on explicitly, here https://github.com/resoai/TileBoard/blob/b7da8bd85b1080dab28fb6f0c851159e9b76ab60/scripts/controllers/main.js#L832 .

I believe this should either be removed or be overridable by an option. I'd prefer the former.

Maybe, the slider `<div>` should then also be the same color as the tile itself, i.e. changing with the entity state. But I don't know how to do that...

akloeckner:
Just to keep a record: #388 might be the first step to start from scratch as discussed in #313

PR #474, opened 2020-10-12
refactor: cache sliders and unify handling of them

akloeckner:
This cleans up slider code and uses the new `cacheInItem` 

BREAKING CHANGE: The order of attrs and def is changed for a few sliders. This should not normally be a problem, because user would only use one of the possibilities. But... just to give note on that.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -814,8 +814,6 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
             max: config.max || attrs.max || defaults.max,
             step: config.step || attrs.step || defaults.step,
             request: config.request || defaults.request,
-            title: config.title || defaults.title,
-            formatValue: config.formatValue || defaults.formatValue,
             curValue: undefined, // current value received from HA
             oldValue: undefined, // last value received from HA
             newValue: undefined, // new value set by the user through the slider
@@ -858,12 +856,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       return initSliderConf(item, entity, config, DEFAULT_VOLUME_SLIDER_OPTIONS);
    };
 
-   $scope.getLightSliderValue = function (conf) {
-      if (conf.formatValue) {
-         return conf.formatValue(conf);
+   $scope.getLightSliderValue = function (slider, conf) {
+      if (slider.formatValue) {
+         return slider.formatValue(conf);
       }
 
-      return conf.value;
+      return conf.getSetValue();
    };
 
    $scope.openLightSliders = function (item, entity) {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -977,16 +977,15 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
 
    const setSliderValue = debounce(setSliderValueFn, 250);
 
-   function setSliderValueFn (item, entity, value) {
-      if (!value.request) {
+   function setSliderValueFn (item, entity, sliderConf) {
+      if (!sliderConf.request) {
          return;
       }
 
-      const conf = value.request;
       const serviceData = {};
-      serviceData[conf.field] = value.newValue;
+      serviceData[sliderConf.request.field] = sliderConf.newValue;
 
-      callService(item, conf.domain, conf.service, serviceData);
+      callService(item, sliderConf.request.domain, sliderConf.request.service, serviceData);
    }
 
    $scope.sliderChanged = function (item, entity, value) {
@@ -1014,18 +1013,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       setSliderValue(item, entity, value);
    };
 
-   $scope.lightSliderChanged = function (slider, item, entity, value) {
-      if (!item._controlsInited) {
-         return;
-      }
-      if (!slider._sliderInited) {
-         return;
-      }
-      if (!item._sliderInited) {
+   $scope.lightSliderChanged = function (item, entity, slider, sliderConf) {
+      if (!item._controlsInited || !item._sliderInited || !slider._sliderInited) {
          return;
       }
 
-      setSliderValue(item, entity, value);
+      setSliderValue(item, entity, sliderConf);
    };
 
    $scope.toggleSwitch = function (item, entity, callback) {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -814,6 +814,8 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
             max: config.max || attrs.max || defaults.max,
             step: config.step || attrs.step || defaults.step,
             request: config.request || defaults.request,
+            title: config.title || defaults.title,
+            formatValue: config.formatValue || defaults.formatValue,
             curValue: undefined, // current value received from HA
             oldValue: undefined, // last value received from HA
             newValue: undefined, // new value set by the user through the slider
@@ -856,12 +858,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       return initSliderConf(item, entity, config, DEFAULT_VOLUME_SLIDER_OPTIONS);
    };
 
-   $scope.getLightSliderValue = function (slider, conf) {
-      if (slider.formatValue) {
-         return slider.formatValue(conf);
+   $scope.getLightSliderValue = function (conf) {
+      if (conf.formatValue) {
+         return conf.formatValue(conf);
       }
 
-      return conf.getSetValue();
+      return conf.value;
    };
 
    $scope.openLightSliders = function (item, entity) {
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -796,8 +796,18 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       return page.header;
    };
 
-   function initSliderConf (item, entity, key, def, attrs, value, default_request) {
+   function initSliderConf (item, entity, def, value) {
+      const key = '_c_slider_' + def.field;
+
       return cacheInItem(item, key, function () {
+         const attrs = entity.attributes || {};
+         const value = +attrs[def.field] || 0;
+         const default_request = {
+            domain: 'input_number',
+            service: 'set_value',
+            field: 'value',
+         };
+
          $timeout(function () {
             item._sliderInited = true;
          }, 50);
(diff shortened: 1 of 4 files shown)


scripts/controllers/main.js
```diff
@@ -796,106 +796,54 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-
-      entity.attributes[key] = {
-         max: attrs.max || def.max || 100,
-         min: attrs.min || def.min || 0,
-         step: attrs.step || def.step || 1,
-         value: value || +entity.state || def.value || 0,
-         request: def.request || {
+      return cacheInItem(item, key, function () {
+         const attrs = entity.attributes || {};
+         const default_request = {
```
rchl, line 804: Use camelCase.

rchl (commented):
On Desktop:
1. Click & hold on the tile to open sliders
2. Click "back" to hide sliders
3. Click the tile to turn off light

It usually doesn't turn off on the first click.
It looks like a new issue. On the other hand, the old (current) code has a problem that sliders sometimes close immediately when trying to open them.

scripts/controllers/main.js
```diff
@@ -796,106 +796,54 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
+         }, 50);
 
-      if (!entity.attributes) {
-         entity.attributes = {};
-      }
-      if (entity.attributes[key]) {
-         return entity.attributes[key];
-      }
+         return {
+            max: def.max || attrs.max || 100,
```
rchl, line 815: Your commit message says:

> prefer def over attrs, as it's user defined

How is `def` user-defined? It seems to be hardcoded in this code rather than user-defined. I would think that our defaults should be a fallback and not the first choice since it might actually not match the entity.

rchl: Or actually, I guess you are correct when it comes to the light sliders but not the volume slider. I suppose the volume slider attributes might not always our default so it would be wrong to prefer our default in that case.

scripts/controllers/main.js
```diff
@@ -796,106 +796,54 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-      const attrs = entity.attributes || {};
-      const value = +attrs[def.field] || 0;
-
-      entity.attributes[key] = {
-         max: attrs.max || def.max || 100,
-         min: attrs.min || def.min || 0,
-         step: attrs.step || def.step || 1,
-         value: value || +entity.state || def.value || 0,
-         request: def.request || {
+      return cacheInItem(item, key, function () {
```
rchl, line 802: Actually, I think that caching here is not safe either. In a situation where the value (for example light brightness) changes externally instead of manually changing the slider then due to caching, we will show sliders with outdated values the next time they are opened.

Values can change externally by for example changing light brightness using different instances of TileBoard (desktop vs. mobile, for example) or using HA itself instead of tileboard. Or the light brightness might change on its own from some automation.

scripts/controllers/main.js
akloeckner: `def` usually comes from user-defined `slider` config. It is only hardcoded for the volume slider, in which the properties for `min`, `max`, `step`, `field` should be ok to be fixed.

I'm not sure though, if `def` stands for default or for defined... So, maybe we should have both...

scripts/controllers/main.js
akloeckner: True again as for the background image... Not sure... Maybe this should simply not be changed then? Or not chached? Or, you never know, maybe the caching within the entity was actually on purpose...

scripts/controllers/main.js
rchl: Why are you assuming that it should be OK to always use our defaults for volume sliders? It would be OK if HA would always normalize volume to 0-1 range but have you checked that it actually does that?

scripts/controllers/main.js
rchl: Caching within an entity might have given an illusion of cache working while, I think, the entity is always re-created on every state change. So not caching here would probably be equivalent to old code. Or well, not entirely since an entity whose state doesn't change wouldn't be re-created but in cases like opening sliders caching is probably not very beneficial anyway (it's not something that happens many times a second).

akloeckner:
> It usually doesn't turn off on the first click.
> It looks like a new issue.

Seems to be a timing issue on your end, which I'd attribute to the `timeout`s in this code around `slider._sliderInited` and `item._sliderInited`. If you go through your procedure fast enough, you might end up being blocked by some of the timeouts. My HomeMatic lights are probably just too slow to run into this problem.

I don't have that here. Instead, the sliders sometimes just don't open, but the light turns on. Which is an old issue,  #312. I see the culprit in this line:
https://github.com/resoai/TileBoard/blob/f9ea1efbdb2f06bf02847266cc3ab1d07ac22a9d/scripts/controllers/main.js#L917

> On the other hand, the old (current) code has a problem that sliders sometimes close immediately when trying to open them.

That's also an old issue, #405. It's caused by "ghost" clicks: Mouse-up generates the `tap` event and just after (I think. Or before?) issues a `click`. So, if you release the mouse just where the back button appears, the button gets clicked right away. I have a rather hands-on fix for that running for quite some time without any issues in my setup. But we wanted to check out the full-blown solution in #407 for some more time. Maybe it's time to PR my fix? :-)

scripts/controllers/main.js
akloeckner: It does: https://www.home-assistant.io/integrations/media_player/#service-media_playervolume_set

> Float for volume level. Range 0..1

But your're right. We'd rather have those volume settings as default. And allow a user-definable input. You never know, what people are up to...

scripts/controllers/main.js
akloeckner: We could just do

```
return cacheInItem(entity, key, function () {
```

That would restore old behavior.

akloeckner: I did resolve to exactly that by using a getterSetter-function.

akloeckner (commented):
That's a whole bunch of additional commits. Let me know, if you prefer to start anew and open another PR. 

Overall (non-breaking) changes now (starting from master):
* unify order of function arguments to be `item, entity, others`
* use getterSetter function for slider value
* store slider default configuration in `globals/constants`
* unify slider configurations in `initSliderConf` function
  * cache config in `item` (instead `entity.anttributes`) to allow multiple items per entity
  * set slider options in order `user config || entity attributes || default from globals`
  * define getterSetter function for current value
* make variables names in `setSlider*` functions mor explicit
* add `z-index` to slider in slider tile, because slider was hidden by value otherwise

BREAKING CHANGES:
(which I all deem more appropriate, if not a bug fix)
* `getSliderConf`: 
  * option precendence inverted from `attributes || user options` to `user options || attributes`.
  * more sources for actual value allowed
* `getLightSliderConf`:
  * default options changed to `max=255` and correct `request`
  * defaulting value to `min` removed. I shall still fix that. Just realized now.

akloeckner:
fixed up the value precedence to match the prior code (excpet additional allowed inputs)

akloeckner:
> It usually doesn't turn off on the first click.
> It looks like a new issue.

Just to be sure and not create new issues knowingly: can you check, if I'm right and it's a timing issue? You should be able to verify by clicking more slowly. ;-) Also, the issue should not manifest starting with the second time you open the sliders.

scripts/controllers/main.js
```diff
@@ -796,106 +796,64 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
    };
 
    $scope.getVolumeConf = function (item, entity) {
-      if (!entity.attributes) {
-         entity.attributes = {};
-      }
-      if (entity.attributes._c) {
-         return entity.attributes._c;
-      }
+      const config = { };
```
rchl, line 850: We can skip creating this variable. :)

scripts/controllers/main.js
```diff
@@ -796,106 +796,64 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
+            oldValue: undefined, // last value received from HA
+            newValue: undefined, // new value set by the user through the slider
+            value: undefined, // the most current value from all the above
+         };
+         sliderConf.getSetValue = function (newValue) {
+            if (arguments.length) {
+               sliderConf.newValue = newValue;
+               sliderConf.value = newValue;
+            } else {
+               sliderConf.curValue = +entity.attributes[config.field] || +entity.attributes[defaults.field] || +entity.state || config.value || defaults.value || config.min || defaults.min || entity.attributes.min || 0;
```
rchl, line 827: [style] Line got really long here. You can extract `entity.attributes` to a separate variable to make it shorter and more readable. And can use restructuring for that: `const { attributes } = entity;`

scripts/controllers/main.js
```diff
@@ -796,106 +796,64 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-         step: attrs.step || def.step || 1,
-         value: value || +entity.state || def.value || 0,
-         request: def.request || {
-            domain: 'input_number',
-            service: 'set_value',
-            field: 'value',
-         },
-      };
+      return cacheInItem(item, key, function () {
+         const attrs = entity.attributes || {};
```
rchl, line 803: [style] I prefer that variables are created as close to being used as possible. So this could be just before `const sliderConf`.

scripts/controllers/main.js
```diff
@@ -1019,55 +977,39 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
 
    const setSliderValue = debounce(setSliderValueFn, 250);
 
-   function setSliderValueFn (item, entity, value) {
-      if (!value.request) {
+   function setSliderValueFn (item, entity, sliderConf) {
+      if (!sliderConf.request) {
```
rchl, line 981: [style] `sliderConf.request` is repeated many times in this function. Can do: `const { request } = sliderConf;`

rchl:
> > It usually doesn't turn off on the first click.
> > It looks like a new issue.
> 
> Just to be sure and not create new issues knowingly: can you check, if I'm right and it's a timing issue? You should be able to verify by clicking more slowly. ;-) Also, the issue should not manifest starting with the second time you open the sliders.

I'm not sure anymore. It seems to work fine now :)

scripts/controllers/main.js
```diff
@@ -1019,55 +977,39 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-      serviceData[conf.field] = value.value;
+      serviceData[sliderConf.request.field] = sliderConf.newValue;
 
-      callService(item, conf.domain, conf.service, serviceData);
+      callService(item, sliderConf.request.domain, sliderConf.request.service, serviceData);
    }
 
-   $scope.sliderChanged = function (item, entity, value) {
+   $scope.sliderChanged = function (item, entity, sliderConf) {
       if (!item._sliderInited) {
```
rchl, line 992: Would it make sense to check `sliderConfig._sliderInited` now instead? Also in places below.

akloeckner: Not sure. Actually, I don't really understand the use of those `inited` flags. But I guess, there is a reason to them. So, I just made sure they are still created. 

Usually, I would assume that code is written such that those functions are only called when `inited` is defined by definition anyway. Especially, that the flags are created with a `timeout`, and even different timeout delays... I don't see the use for that. 

Well, at least for volume, your suggestion won't work, because the slider configuration is just `{}` and not stored anywhere... Also, what is being passed here, is the `item._c_slider_...` cached configuration, not the `item.slider` user-defined configuration. So, the fact that `sliderConf != undefined` already is a sign for the slider being inited.

scripts/controllers/main.js
akloeckner: Learning new things every time...

scripts/controllers/main.js
```diff
@@ -978,14 +977,16 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    const setSliderValue = debounce(setSliderValueFn, 250);
 
    function setSliderValueFn (item, entity, sliderConf) {
-      if (!sliderConf.request) {
+      const { request, newValue } = sliderConf;
```
rchl, line 980: I wouldn't create a separate variable for properties that you are only checking once.

scripts/controllers/main.js
rchl: OK

rchl:
Love the changes.
And sorry for picking too picky at times. :)

scripts/controllers/main.js
akloeckner: Just to note: More places, where I wouldn't expect the present checks:
https://github.com/resoai/TileBoard/blob/31d5a5ae6440ef2ccac6d3c977133216fef3e9cd/scripts/controllers/main.js#L883
https://github.com/resoai/TileBoard/blob/31d5a5ae6440ef2ccac6d3c977133216fef3e9cd/scripts/controllers/main.js#L1015

PR #488, opened 2020-10-18
fix(LIGHT): show color picker based on feature check

rchl:
This is a behavior change since now even if the user doesn't define
sliders and the light supports color change, then picker will still show up
on long press.

We maybe should leave support for "item.colorpicker" to be able to
override it to "false" but not sure if it's worth it.

@akloeckner

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,10 +356,10 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    /** type: LIGHT **/
    /* sliders: list of slider object. See slider documentation below */
    sliders: [{}],
-   /* colorpicker: whether or not the color picker should be shown. 
+   /* colorpicker: whether or not the color picker should be shown.
     * Valid options: true, false, function. Default: auto-detected
     */
-   colorpicker: true,
+   colorpicker: false,
    /** type: POPUP_IFRAME **/
    url: String || Function,
    /* optional */
--- a/README.md
+++ b/README.md
@@ -356,7 +356,10 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    /** type: LIGHT **/
    /* sliders: list of slider object. See slider documentation below */
    sliders: [{}],
-   
+   /* colorpicker: whether or not the color picker should be shown. 
+    * Valid options: true, false, function. Default: auto-detected
+    */
+   colorpicker: true,
    /** type: POPUP_IFRAME **/
    url: String || Function,
    /* optional */
--- a/README.md
+++ b/README.md
@@ -357,11 +357,6 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    /* sliders: list of slider object. See slider documentation below */
    sliders: [{}],
    
-   /* colorpicker: whether or not the color picker should be used. 
-    * Only works with lights that have the rgb_color attribute 
-    * Valid options: true, false 
-    */
-   colorpicker: true,
    /** type: POPUP_IFRAME **/
    url: String || Function,
    /* optional */
--- a/scripts/controllers/main-utilities.js
+++ b/scripts/controllers/main-utilities.js
@@ -1,0 +1,41 @@
+import angular from 'angular';
+import { TILE_DEFAULTS, TYPES } from '../globals/constants';
+
+export function mergeConfigDefaults (pages) {
+   for (const page of pages) {
+      for (const group of page.groups) {
+         mergeTileListDefaults(group.items);
+      }
+   }
+   return pages;
+}
+
+function mergeTileListDefaults (tiles) {
+   if (!Array.isArray(tiles)) {
+      return;
+   }
+   for (const [index, tile] of tiles.entries()) {
+      if (tile.type in TILE_DEFAULTS) {
+         tiles[index] = mergeTileDefaults(tile);
+      }
+      switch (tile.type) {
+         case TYPES.CAMERA_THUMBNAIL:
+            tile.fullscreen = mergeTileDefaults(tile.fullscreen);
+            break;
+         case TYPES.DOOR_ENTRY:
+            if (tile.layout?.camera) {
+               tile.layout.camera = mergeTileDefaults(tile.layout.camera);
+            }
+            mergeTileListDefaults(tile.layout?.tiles);
+            break;
+         case TYPES.POPUP:
+            mergeTileListDefaults(tile.popup?.items);
+            break;
+      }
+   }
+   return tiles;
+}
+
+function mergeTileDefaults (tile) {
+   return angular.merge({}, TILE_DEFAULTS[tile.type], tile);
+}
--- a/scripts/controllers/main-utilities.js
+++ b/scripts/controllers/main-utilities.js
@@ -15,10 +15,10 @@ function mergeTileListDefaults (tiles) {
       return;
    }
    for (const [index, tile] of tiles.entries()) {
-      if (tile.type in TILE_DEFAULTS) {
-         tiles[index] = mergeTileDefaults(tile);
-      }
+      tiles[index] = mergeTileDefaults(tile);
       switch (tile.type) {
+         case TYPES.CAMERA:
+         case TYPES.CAMERA_STREAM:
          case TYPES.CAMERA_THUMBNAIL:
             tile.fullscreen = mergeTileDefaults(tile.fullscreen);
             break;
@@ -37,5 +37,8 @@ function mergeTileListDefaults (tiles) {
 }
 
 function mergeTileDefaults (tile) {
-   return angular.merge({}, TILE_DEFAULTS[tile.type], tile);
+   if (tile && tile.type in TILE_DEFAULTS) {
+      return angular.merge({}, TILE_DEFAULTS[tile.type], tile);
+   }
+   return tile;
 }
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -919,7 +919,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.getGaugeField = function (field, item, entity) {
-      return parseFieldValue(item.settings[field], item, entity);
+      return getItemFieldValue(`settings.${field}`, item, entity);
    };
 
    $scope.itemURL = function (item, entity) {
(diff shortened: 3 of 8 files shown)


akloeckner:
I believe, we discussed in #313 that we wanted to make an automatic history/brightness/temperature/color popup for this.

However, I actually like the slim looks of the light sliders on the tile. It's just not usable for me, since it turns on the light on full brightness before I even get a chance to adjust the brightness (see #312).

So, in general, I believe we should revisit the whole light slider thing. Either by a popup. Or by changed long-press behavior plus automatic slider configuration. I just wonder where we'd put a history graph then, if we want it to be available for lights.

That being said, more automatic configurations are a good thing. Still, the major point TileBoard makes is its being customizable. So, I'd prefer to allow overriding automatic choices, here with `colorpicker: false`. Plus, if we automate the colorpicker, we should certainly also automate brightness. And possibly color temperature, too.

Do you maybe see a way to have overrideable default choices? E.g., don't check `item.colorpicker`, but check `itemField('colorpicker', item, entity)`. Then, merge in a `DEFAULT_SETTINGS.LIGHT` here (or any other place you see more fit):

https://github.com/resoai/TileBoard/blob/e796af56a7918745e8f20bd49b0325db20f231ee/scripts/controllers/main.js#L2152

with 
```
DEFAULT.LIGHT = { 
   ...
   colorpicker = (item, entity) => supportsFeature(FEATURES.LIGHT.COLOR, entity)
   ...};
```

Would that make sense? We could then easily add lots of default options. And make all of them overrideable...

rchl:
I've done as you've suggested with the exception that I'm not merging defaults dynamically from `getItemFieldValue` but initially on loading app. That won't introduce any runtime overhead.

akloeckner:
Very nice new capacity! I'll need some time to check it out. But I'm excited!

README.md
```diff
@@ -356,10 +356,8 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
    /** type: LIGHT **/
    /* sliders: list of slider object. See slider documentation below */
    sliders: [{}],
-   
-   /* colorpicker: whether or not the color picker should be used. 
-    * Only works with lights that have the rgb_color attribute 
-    * Valid options: true, false 
+   /* colorpicker: whether or not the color picker should be shown.
```
akloeckner, line 359: Trailing space

scripts/controllers/main-utilities.js
```diff
@@ -0,0 +1,41 @@
+import angular from 'angular';
```
akloeckner, line 1: Is there a reason that this is a new file? Why not put it in existing utils?

scripts/controllers/main.js
```diff
@@ -902,58 +905,21 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-      if (item.settings && field in item.settings) {
-         return parseFieldValue(item.settings[field], item, entity);
-      }
-
-      if (field in GAUGE_DEFAULTS) {
-         return parseFieldValue(GAUGE_DEFAULTS[field], item, entity);
-      }
-
-      return null;
+      return parseFieldValue(item.settings[field], item, entity);
```
akloeckner, line 922: We could use `getItemFieldValue` here. It feels a bit odd to not check the existence of `settings`. Although it's clearly there by defaults...

scripts/globals/constants.js
```diff
@@ -269,3 +276,25 @@ export const DEFAULT_VOLUME_SLIDER_OPTIONS = {
       field: 'volume_level',
    },
 };
+
+export const TILE_DEFAULTS = {
+   [TYPES.GAUGE]: {
+      settings: {
+         backgroundColor: 'rgba(0, 0, 0, 0.1)',
+         foregroundColor: 'rgba(0, 150, 136, 1)',
+         size (item) {
```
akloeckner, line 285: What does this line do? Is it equivalent to the former property definition?

scripts/globals/constants.js
```diff
@@ -269,3 +276,25 @@ export const DEFAULT_VOLUME_SLIDER_OPTIONS = {
...
+         thick: 6,
+         type: 'full',
+         min: 0,
+         max: 100,
+         cap: 'butt',
+         thresholds: {},
+      },
+   },
+   [TYPES.LIGHT]: {
+      colorpicker: (item, entity) => supportsFeature(FEATURES.LIGHT.COLOR, entity),
```
akloeckner, line 298: * Should we also add at least the brightness slider? 
* Actually, the color picker is too much for my tiny tiles already. But we can override the defaults, so it's fine.

scripts/globals/constants.js
```diff
@@ -269,3 +276,25 @@ export const DEFAULT_VOLUME_SLIDER_OPTIONS = {
       field: 'volume_level',
    },
 };
+
+export const TILE_DEFAULTS = {
```
akloeckner, line 280: Should we explain somewhere in the docs that the following will now be possible in `config.js`?
```
TILE_DEFAULTS[TYPES.LIGHT].colorpicker = false;
```
Maybe be even add a line to the example:
```
TILE_DEFAULTS = angular.merge(TILE_DEFAULTS, {
   // override defaults here, e.g.:
   [TYPES.LIGHT]: {
      colorpicker: false
   }
});
```

scripts/globals/utils.js
```diff
@@ -129,3 +129,13 @@ export const toAbsoluteServerURL = function (path) {
...
 };
+
+export function supportsFeature (feature, entity) {
+   if (!('supported_features' in entity.attributes)) {
+      return false;
+   }
+
+   const features = entity.attributes.supported_features;
+
+   return (features | feature) === features;
```
akloeckner, line 140: Can we abbreviate? 
```
(feature & feature) !== 0
```

Then we could also spare the constant.

akloeckner (commented):
Looks great, just a few comments that might help improve even more.

scripts/controllers/main-utilities.js
```diff
@@ -0,0 +1,41 @@
...
+function mergeTileListDefaults (tiles) {
+   if (!Array.isArray(tiles)) {
+      return;
+   }
+   for (const [index, tile] of tiles.entries()) {
+      if (tile.type in TILE_DEFAULTS) {
+         tiles[index] = mergeTileDefaults(tile);
+      }
+      switch (tile.type) {
+         case TYPES.CAMERA_THUMBNAIL:
```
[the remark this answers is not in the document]
rchl: > * Wouldn't we need the other camera types here, too?

Most likely yes. I've been basing this on TILE_EXAMPLE.md which doesn't list all possible combinations.

> * Also, the `fullscreen` could possibly have different defaults based on its own `type`, couldn't it. So, the `if (tile.type in TILE_DEFAULTS)` should maybe be moved to `mergeTileDefaults`?

I'll have to investigate what is possible to use there.

> * Finally, `fullscreen` might not be set. Not sure, if that's a problem.

The code should handle that. Basically will overwrite `undefined` with `undefined`.

scripts/controllers/main-utilities.js
rchl: Existing utils are for public use and are also exposed on `window`. This one is only for internal use.

scripts/controllers/main.js
rchl: Sure

scripts/globals/constants.js
rchl: Yes, it's a shorthand for defining a function property on an object.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

scripts/globals/utils.js
rchl: I think that makes more sense

scripts/globals/constants.js
rchl: We should but I haven't figured out a way to state that in a generic way. Maybe we should wait until we have all tiles covered that way because now the instructions would have to be rather specific.

scripts/controllers/main-utilities.js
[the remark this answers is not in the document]
resoai: `CAMERA_THUMBNAIL` should be deprecated and removed at some point as HA had deprecated it a while ago.

scripts/controllers/main-utilities.js
akloeckner: Ok.

scripts/controllers/main-utilities.js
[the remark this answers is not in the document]
akloeckner: > I'll have to investigate what is possible to use there.

I believe basically `fullscreen` defines another camera tile. And taking @resoai's comment into account, there are only `CAMERA` and `CAMERA_STREAM`.

scripts/globals/constants.js
akloeckner: Also fine with me, that would give us time to sort out last issues with this new scheme before people start to use it. I'll certainly help find those issues, because I'll switch my whole setup to use the defaults. 

We just shouldn't forget...

README.md
```diff
@@ -356,10 +356,8 @@ Tile Object. [Click here for some real-life examples](TILE_EXAMPLES.md)
...
    /* sliders: list of slider object. See slider documentation below */
    sliders: [{}],
-   
-   /* colorpicker: whether or not the color picker should be used. 
-    * Only works with lights that have the rgb_color attribute 
-    * Valid options: true, false 
+   /* colorpicker: whether or not the color picker should be shown. 
+    * Valid options: true, false, function. Default: auto-detected
     */
    colorpicker: true,
```
akloeckner, line 362: Maybe, use `false` as example? After all, that will be the use case: removing the picker, if undesired. We wouldn't add a picker for an entity that has no color support...

scripts/controllers/main-utilities.js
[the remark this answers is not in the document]
rchl: I'll add handling for `CAMERA` and `CAMERA_STREAM` and fix the issue with code crashing on undefined `item.fullscreen`. But will leave deprecating `CAMERA_THUMBNAIL` for later.

scripts/globals/constants.js
rchl: I'd like to add handling for other sliders separately as that would be more involved change.

akloeckner: Alright!

akloeckner:
No further comments. This is a great improvement!

PR #489, opened 2020-10-18
fix: long-press closes popups and light sliders immediately

### LINKED ISSUE #405: popup closing immediately

akloeckner:
fixes #405
closes #407

Thought, I'd share this. It has been running here for quite some time and fixes some annoying issues. I know, it is not as full-blown as #407. But it works. And #407 hasn't seen much progress lately. The down-side of this fix is, that we *really* cannot click in the region of a long-press for some fixed timeout. 

using the fix from here: https://gist.github.com/jtangelder/361052976f044200ea17
as suggested by hammer.js here: https://hammerjs.github.io/tips/


rchl:
Just cleaned up a little and rebased to master (as otherwise, popup was failing to open).

I'm fine with this one. :)

PR #508, opened 2020-10-29
fix: fullsize popup cropped at the bottom

### LINKED ISSUE #423: History popup cropped on bottom
I found that my popups are cropped at the bottom, when their height hits the max-height:

Screenshot_2020-09-07-22-29-36

It can be fixed by subtracting 60px in custom css:
```
.popup--tiles {
  max-height: calc(100vh - 50px - 80px - 30px - 62px); 
  border: 1px solid red;
}
```

It's easiest testable by opening a history popup on mobile and turning your phone to landscape. Both in Chrome and Firefox. I am pretty sure, this was correct when I migrated history popup to the modular popup.

Do you see this, too? Or is it some strange setting of mine, which I just didn't find? 

If it's a real issue: why 60px?!

akloeckner:
fixes #423

this uses the first approach suggested in the issue.
I just couldn't get the other one to work...

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -72,6 +72,13 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       }
    }
 
+   function innerHeightToCSS () {
+      const vh = window.innerHeight * 0.01;
+      document.documentElement.style.setProperty('--vh', `${vh}px`);
+   }
+   innerHeightToCSS();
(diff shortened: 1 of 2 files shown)


rchl:
Seems fine to me.

In the future, we could probably just switch to flexbox which should make it easy to do without using `vh` units. flexbox has even better compatibility than `vh` unit as far as I can see on "caniuse".

PR #510, opened 2020-10-29
fix: unify examples

akloeckner:
I started to play around with unified examples and I believe this is as far as I can get. It's far from complete, but I tried to line out the general structure. The code fences and block comments are not pretty, but it's the least invasive I could think of.

What do you think? Maybe you have ideas how to generate a markdown view for Github display using rollup? 

I thought, I'd open this draft PR as a discussion place...

Here's the rendered file: https://github.com/akloeckner/TileBoard/blob/test/unify-examples/config.example.js

### DIFF
--- a/config.example.js
+++ b/config.example.js
@@ -24,7 +24,7 @@ var CONFIG = {
 ## Connection <sup>[back to toc](#table-of-contents)</sup>
 
 ```js
-/*
+*/
    serverUrl: 'http://' + location.hostname + ':8123',
    wsUrl: 'ws://' + location.hostname + ':8123/api/websocket',
    debug: false, // Prints entities and state change info to the console.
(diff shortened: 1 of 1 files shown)


config.example.js
```diff
@@ -24,7 +24,7 @@ var CONFIG = {
 ## Connection <sup>[back to toc](#table-of-contents)</sup>
 
 ```js
-/*
+*/
```
rchl, line 27: I'm really skeptical about this weird format that you've applied here.

A JS file with a weird mix of markdown embedded in block comments.

I would think that normal, nicely formatted block comments would be perfectly fine...

rchl:
> The code fences and block comments are not pretty, but it's the least invasive I could think of.

It's a very bizarre looking solution that is neither looking nicely (the source code, I mean) nor is easy to edit.

If the intention was to create something that can be used both as a default config and as a documentation page then maybe that's not necessary. We could just have a very simple example config and a separate markdown documentation page that documents everything. I think that should be good enough.

So I don't we should be investing time into this format nor investigate whether we could do something with post-processing with rollup. I suppose we could do anything we want if we use some templating language (like ejs) but it won't necessarily make it easier for us.

akloeckner:
> If the intention was to create something that can be used both as a default config and as a documentation page

It was. :-) And it was good to discuss before investing too much time. 

A very simple example, say `sun.sun`, some welcome text-list, and a link (or popup iframe) to the (one!) documentation on Github indeed is probably the best (and simplest) solution...

PR #511, opened 2020-10-29
fix(HOMEKIT): make colors of ON tiles consistent

### LINKED ISSUE #448: Some icons don't have their background applied correctly?
Hello,

i'm currently setting up a dashboard with Tileboard and - even though the configuration is quite... elaborate.. - like what i could create so far.

When using customTheme: CUSTOM_THEMES.HOMEKIT i noticed that two buttons don't have their background applied correctly (at least not the same as the other icons) - see this screenshot:

Bildschirmfoto 2020-09-30 um 15 26 21

In the second group (Haustür & Alarm) both the "Haustür" (type: TYPES.LOCK) and "Briefkasten" (type: TYPES.SENSOR.ICON) don't have their background applied correctly.

I can workaround this with customStyles i guess ('background': 'linear-gradient(to right bottom, rgba(253, 251, 249, 0.95), rgba(245, 245, 245, 0.8))'), however i'd prefer a "real" fix 😄 

I've also updated to 1.0.3, however that didn't seem to help so far.

I hope i've given you a few pointers to fix that - if you need anything else, don't hesitate to ask!

Thank you for Tileboard!

P.S.: not sure if i should post this in a separate issue, however i'd _really_ like to align the second and third group below each other. I know, from reading the community board thread, that you're seeing issues regarding the grid layout. Of course, that is valid i think - however, if one would sacrifice the space of one tile for the title, everything should be aligned well?

rchl:
Resolves #448

### DIFF
--- a/styles/themes.less
+++ b/styles/themes.less
@@ -451,11 +451,15 @@
   color: #fff;
   background: #eee !important;
 
+  @TILE_COLOR_ON: #fff;
+  @TILE_BACKGROUND_ON: rgb(255, 255, 255);
+  @TILE_COLOR_OFF: rgb(84, 77, 75);
+  @TILE_BACKGROUND_OFF: linear-gradient(to right bottom, rgba(253, 251, 249, 0.95), rgba(245, 245, 245, 0.8));
+
   .item {
     color: #363636;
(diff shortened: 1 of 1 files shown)


resoai:
Thanks!

PR #516, opened 2020-11-02
Linting improvement on committing

rchl:
fix(dev): don't lint unstaged changes on commit
fix(dev): automatically fix eslint errors in changed files on commit

@akloeckner

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,9 +18,14 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn lint",
+      "pre-commit": "lint-staged",
       "post-merge": "node post-merge-warn.js"
     }
+  },
+  "lint-staged": {
+    "*.js": [
+      "npm run lint"
+    ]
   },
   "repository": {
     "type": "git",
@@ -46,6 +51,7 @@
     "eslint": "^7.12.1",
     "husky": "^4.3.0",
     "less": "^3.12.2",
+    "lint-staged": "^10.5.1",
     "print-message": "^3.0.1",
     "rollup": "^2.33.1",
     "rollup-plugin-cpy": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   },
   "lint-staged": {
     "*.js": [
-      "npm run lint"
+      "eslint --cache --fix",
+      "yarn lint"
     ]
   },
   "repository": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1116,6 +1116,13 @@ ansi-colors@^4.1.1:
   resolved "https://registry.yarnpkg.com/ansi-colors/-/ansi-colors-4.1.1.tgz#cbb9ae256bf750af1eab344f229aa27fe94ba348"
   integrity sha512-JoX0apGbHaUJBNl6yF+p6JAFYZ666/hhCGKN5t9QFjbJQKUU/g8MNbFDbvfrgKXvI1QpZplPOnwIo99lX/AAmA==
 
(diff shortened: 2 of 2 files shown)


akloeckner:
I checked out the branch, did two eslint-offending modifications, staged one of those, and tried to commit. I got the following error:

```
homeassistant@raspberrypi:~/.homeassistant/www/tileboard $ git diff --cached
diff --git a/scripts/controllers/main.js b/scripts/controllers/main.js
index 9f53c77..d4208e3 100755
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1605,7 +1605,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
                      });
                   }));
             } else {
-               entityId.forEach(function (entityId, index) {
+               entityId.forEach(function (entityId,index) {
                   historyObject.watchers.push($scope.$watch(
                      function () {
                         return $scope.states[entityId].state;
homeassistant@raspberrypi:~/.homeassistant/www/tileboard $ git diff
diff --git a/scripts/controllers/main.js b/scripts/controllers/main.js
index d4208e3..5d1fb83 100755
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1630,7 +1630,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       return item[key];
    }

-   $scope.initTileHistory = function (item, entity) {
+   $scope.initTileHistory = function (item,entity) {
       return cacheInItem(item, '_historyObject', () => getHistoryObject(item, entity, item));
    };

homeassistant@raspberrypi:~/.homeassistant/www/tileboard $ git commit
husky > pre-commit (node v14.5.0)
✔ Preparing...
✔ Hiding unstaged changes to partially staged files...
✔ Running tasks...
✖ Prevented an empty git commit!
↓ ✖ lint-staged failed due to a git error. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up...

  ⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration

husky > pre-commit hook failed (add --no-verify to bypass)
```

(I'd expect a nano to open up to let me type my commit message...)

akloeckner:
Also: Why do we run `eslint --fix` **and** `yarn lint`?

rchl:
@akloeckner You've staged a change with only a linting issue. The script on commit has automatically fixed the issue in your staged file and you were left with nothing to commit. So the behavior is correct.

> Also: Why do we run `eslint --fix` **and** `yarn lint`?

The first one automatically fixed any issues in staged files and then the second one checks the whole code-base for linting issues. The reason for the second is:
 - eslint is not able to fix all issues it reports
 - your change might have affected linting in other files (for example when changing eslint configuration itself)

akloeckner:
I see. Checked again and it works for me! :+1:

akloeckner:
I see. Checked again and it works for me! :+1:

One more question: do we reallywant to add the changed files again? Or would it make sense to leave that up to the user? (That's how HA core code behaves.) Generally, I'm skeptical about software that thinks, it's smarter than me...

akloeckner:
:-) ok. Too late...

rchl:
I think I don't see any issue with adding automatically. These are fixes you the user has to add and commit anyway if he wants to make a commit so why not spare extra steps? Would there likely be a case when the user doesn't want to do that?

akloeckner:
At least, we don't have a possibility to check the changes made anymore. But if you trust the software enough to let it make such unsupervised changes, I'll try and follow your lead. :-) I'll just open another issue, if it breaks something.

PR #519, opened 2020-11-05
fix: improve handling of merging defaults for popup tiles

akloeckner:
This will let the user overwrite default popup layouts. And it makes the code more readable.

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1,6 +1,6 @@
 import angular from 'angular';
 import Hammer from 'hammerjs';
-import { mergeConfigDefaults } from './main-utilities';
+import { mergeConfigDefaults, mergeTileConfigs } from './main-utilities';
 import { App } from '../app';
 import { TYPES, FEATURES, HEADER_ITEMS, MENU_POSITIONS, GROUP_ALIGNS, TRANSITIONS, MAPBOX_MAP, YANDEX_MAP, DEFAULT_SLIDER_OPTIONS, DEFAULT_LIGHT_SLIDER_OPTIONS, DEFAULT_VOLUME_SLIDER_OPTIONS, DEFAULT_POPUP_HISTORY, DEFAULT_POPUP_IFRAME, DEFAULT_POPUP_DOOR_ENTRY } from '../globals/constants';
 import { debounce, leadZero, supportsFeature, toAbsoluteServerURL } from '../globals/utils';
@@ -1601,7 +1601,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.openPopupHistory = function (item, entity) {
-      const layout = cacheInItem(item, '_popupHistory', () => angular.merge(DEFAULT_POPUP_HISTORY(item, entity), {
+      const layout = cacheInItem(item, '_popupHistory', () => mergeTileConfigs({}, DEFAULT_POPUP_HISTORY(item, entity), {
          classes: getItemFieldValue('history.classes', item, entity) || [],
          items: [getItemFieldValue('history', item, entity) || {}],
       }));
@@ -1609,7 +1609,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.openPopupIframe = function (item, entity) {
-      const layout = cacheInItem(item, '_popupIframe', () => angular.merge(DEFAULT_POPUP_IFRAME(item, entity), {
+      const layout = cacheInItem(item, '_popupIframe', () => mergeTileConfigs({}, DEFAULT_POPUP_IFRAME(item, entity), {
          classes: getItemFieldValue('iframeClasses', item, entity) || [],
          items: [{
             customStyles: getItemFieldValue('iframeStyles', item, entity) || {},
@@ -1619,7 +1619,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.openDoorEntry = function (item, entity) {
-      const layout = cacheInItem(item, '_popupDoorEntry', () => angular.merge(DEFAULT_POPUP_DOOR_ENTRY(item, entity), {
+      const layout = cacheInItem(item, '_popupDoorEntry', () => mergeTileConfigs({}, DEFAULT_POPUP_DOOR_ENTRY(item, entity), {
          items: [
             getItemFieldValue('layout.camera', item, entity) || {},
             ...(getItemFieldValue('layout.tiles', item, entity) || []),
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -2,7 +2,7 @@ import angular from 'angular';
 import Hammer from 'hammerjs';
 import { mergeConfigDefaults } from './main-utilities';
 import { App } from '../app';
-import { TYPES, FEATURES, HEADER_ITEMS, MENU_POSITIONS, GROUP_ALIGNS, TRANSITIONS, MAPBOX_MAP, YANDEX_MAP, DEFAULT_SLIDER_OPTIONS, DEFAULT_LIGHT_SLIDER_OPTIONS, DEFAULT_VOLUME_SLIDER_OPTIONS } from '../globals/constants';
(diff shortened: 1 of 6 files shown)


scripts/controllers/main.js
```diff
@@ -1635,70 +1635,30 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
-            classes: ['-item-fullsize'],
-            customStyles: {
-               width: null,
-               height: null,
-               top: null,
-               left: null,
-            },
-         }, getItemFieldValue('history', item, entity))],
+      const layout = cacheInItem(item, '_popupHistory', () => angular.merge(DEFAULT_POPUP_HISTORY(item, entity), {
+         classes: getItemFieldValue('history.classes', item, entity) || [],
```
rchl, line 1639: It doesn't look like `angular.merge` will merge those array values:

```js
> angular.merge({ a: [1] }, { a: [2] }).a
> [2]
```

rchl: I think https://www.npmjs.com/package/lodash.merge should work. `angular.merge` is deprecated anyway and they recommend `lodash.merge`.

scripts/globals/constants.js
```diff
@@ -277,6 +277,65 @@ export const DEFAULT_VOLUME_SLIDER_OPTIONS = {
...
+
+export const DEFAULT_POPUP_DOOR_ENTRY = (item, entity) => ({
+   classes: ['-popup-fullsize'],
+   styles: {},
+   items: [{
+      state: false,
+      title: false,
+      position: [0, 0],
+      action: function (item, entity) {},
+      secondaryAction: function (item, entity) {},
```
rchl, line 328: I suppose we could skip those? Or is there a reason why we've started adding those?

scripts/controllers/main.js
akloeckner: The question for this line might be: What is actually expected? `angular.merge` will merge by `key`, which is `1,2,3,...` in the case of arrays. So, values in the same place will be overwritten, values in other places will be added. Since both use cases should be possible, I'd say this behavior is reasonable to expect.

Here's a more complete test case:

grafik

If we wanted only *add* user-defined classes, we could do so like this:
```
classes: [ , ...(getItemFieldValue('history.classes', item, entity) || [])],
```

scripts/globals/constants.js
akloeckner: I had added `action` for the door entry already. 

Now, I experienced that long-pressing the camera will open a history popup. And long-pressing the history popup will open another history popup. In user-defined popups this might be desirable. But in the default one, I think, we should try to prevent those "chains"...

scripts/controllers/main.js
rchl: In the case of classes, I think the default behavior should be to append user classes to the ones that are there already because then the user doesn't need the knowledge of internal classes to add something. The `[ , ...]` also requires knowledge of how many internal classes are already applied and is error-prone in case something changes internally in the future.

akloeckner: I meant, we could  use the `[, ...]` in `main.js`. The users should definitely not need to to this themselves.

If we were to use another library, I'll need help to include it in the bundle. Just a hint: The merge of arrays will need to use the current (angular) behavior in a few prominent places, e.g., the main tile in popups (i.e. camera, iframe, history)...

In summary, I suggest to change the `classes` line(s) in `main.js` in the `[, ...]` way.

rchl: This seems like an error-prone solution since when changing the defaults we'd need to remember to also update this place and it's not obvious at all.

akloeckner: Hmm... what would you suggest? We could place a reminder comment in the defaults? Add another line to merge classes separately? Use the other library? 

Hint: I had a quick look at the documentation and it seems, lodash behaves the same as angular...
https://lodash.com/docs#merge

rchl: I think you are right about the lodash.merge.

I need to think about it more when I'm less occupied. I think comments are not great.

akloeckner: Just came to my mind: `mergeTileDefaults` should have the same issue. So, maybe define it there once and for all. And simply invoke here.

scripts/globals/constants.js
resoai: Maybe we could do something like `secondaryAction: false`?

akloeckner: I would have preferred that. But it's not possible in the current codebase. Another (maybe even cleaner) idea in that direction would be to make more use of the new tile defaults: Set the default callback there, and simply overwrite it with `undefined` (or anything else `falsy`) in the popup defaults. That might be a bit more involved change, which I would prefer to PR separately, and preferedly after the present PR.

akloeckner: I liked to idea so much, I wanted it done... :-) 
see #522

scripts/controllers/main.js
akloeckner: I played around a bit. This seems to be a good candidate to use: https://lodash.com/docs#mergeWith .

However, it's not as simple as `if 'classes' return concat`, because 
1. angularJS allows strings, objects and arrays for `ngClass`
2. we allow functions and templates in addition.

The former can be fixed by throwing everything in arrays and concatenating those. (See https://docs.angularjs.org/api/ng/directive/ngClass#!)
The latter will likely not work in that manner...

Here's my work in progress: https://github.com/akloeckner/TileBoard/tree/test/merge-tiles

scripts/globals/constants.js
akloeckner: done as @resoai suggested

scripts/controllers/main.js
akloeckner: I thought about it a little while longer, and now I believe that functions and templates are not meaningfully mergeable. (The function's outcome is unpredictable at `mergeTileDefaults` time. And a template is basically a function.) So, if either of the merged properties is a `function` or a `string` we should definitely overwrite the property. If both are arrays, we can append them. 

For the user, that would mean: Supply an array and it will be appended. Supply anthing else and it will override defaults. Still not great and simple... 

The remaining option would be to simply override `classes` *always*. That would be clear at least, but not really a "merge".

akloeckner: An additional thought on that is: we have two kinds of places where we would want to merge tile configurations
a) before "evaluation" of functions and templates, prominent example is during config loading.
b) after "evaluation" of functions and templates, prominent example *might* be the popups.

The a) case could be solved by nesting functions:
```
classes = newFunction(item) { objA.concat( oldFunction(item) ) }
```

Question is: Which `item` is meant in each of the functions? During config loading it will easy, because it can only be the same `item`. In the case of popups, it's not that easy. The function in the default will probably want to use the `item` in the popup. The function in the calling `item` will probably want to use that calling `item`.

So, I'm under the impression, it might boild down to the point: when do we actually want to evaluate functions and templates?

(Sorry for turning into a philosopher. :-) Maybe I should wait for you to be less ocupied...)

akloeckner: I updated https://github.com/akloeckner/TileBoard/compare/feature/defaults...akloeckner:test/merge-tiles in a way that I believe should handle all cases well. (It's untested, so treat it as a sketch.)

* `classes` will always be concatenated.
* to that end, everything that is not an array is embedded in an array
* in order to allow for functions and templates, I wrap everything in another anonymous function and call `this.parseFieleValue` on both properties to be merged
* functions and templates are thus evaluated against the same `item` always
* *unless* we explicitly pass differently evaluated properties to the merger, such as in the line of this discussion. So, we are in total control of what is evaluated when and against which `item`.

Possible downsides:
* everything is `parseFieldValue`ed. But that should be ok, because non-parseable properties are simply piped through.
* properties might become parsed multiple times. But that should be ok for the same reason.

rchl: Didn't mean to leave you here talking to yourself but I don't have much capacity for it currently.

I guess go ahead with your plan and I'll likely approve after a brief look :)

akloeckner: No problem. I prefer to put down my notes, so you could pick up the line of thought. I'll give it another round of thinking and then go ahead.

akloeckner: I added this approach to the PR. Would you mind a quick look? ;-)

PR #521, opened 2020-11-09
feat: use moment.js for timeAgo

akloeckner:
also export the moment class to user space
this will enable the user to do
```
moment.locale('de');
```
in their `config.js`.

### DIFF
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -1,8 +1,5 @@
 import angular from 'angular';
-// @ts-ignore: No type definitions in imported module
-import moment from 'moment/min/moment-with-locales';
-
-export { moment };
+import moment from 'moment';
 
 export const mergeObjects = angular.merge;
 
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -1,4 +1,7 @@
 import angular from 'angular';
+import moment from 'moment/min/moment-with-locales';
+
+export { moment };
 
 export const mergeObjects = angular.merge;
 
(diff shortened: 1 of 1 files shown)


scripts/globals/utils.js
```diff
@@ -1,4 +1,7 @@
 import angular from 'angular';
+import moment from 'moment/min/moment-with-locales';
```
rchl, line 2: Maybe you don't see that but this weird import doesn't have type definitions so it's highlighted as an error.
Could add:
```
// @ts-ignore
```
above it to silence it.

rchl (approved):
This adds quite a chunk of data to the bundle (maybe 500kb or so) but I guess it's probably fine.

scripts/globals/utils.js
akloeckner: I don't see this. But I added the comment. According to the web, it's customary to add a description of the silenced error. I hope, this still works now.

akloeckner:
I tried to find a way to avoid bundling all the locales. We could only bundle "popular" locales like ru, de, en, pl. Or provide info on how the user can load locales. Both don't seem very appealing to me...

akloeckner:
Actually, `timeAgo`  seems to be used only for the default tooltip in history tiles and the corresponding example. Maybe it's better to not touch that function or remove it altogether...

scripts/globals/utils.js
resoai: It would be nice to think about locales for Angular, perhaps.

akloeckner: You mean sth like that?
https://www.npmjs.com/package/angular-timeago

I guess, the footprint will be much smaller. It supports de, en, pl, but not (yet) russian. Angular itself doesn't seem to have a localized `timeAgo` module...

resoai: No, I mean locales for Angular in general as it looks like a lot of people want to localise TB into their language. Having all localities would be a an overkill, but perhaps de, es, fr, ru, pl would be nice to have out of the box.

akloeckner: If I understand the docs right, localizing angular is simply a matter of including the angular localization file, e.g. `i18n/angular-locale_de-de.js`.
https://docs.angularjs.org/guide/i18n

Similar things seem to hold for moment locales, e.g. `locale/fr.js`. Plus we need to call `moment.locale('fr')`.
https://momentjs.com/docs/#/i18n/loading-into-browser/

So, we could
* provide `cat`'ed locales for both modules in some `locales` folder
* include the `moment.locale` call in those files
* provide a `utils.js` function to load those files from `config.js`, e.g. this: https://github.com/resoai/TileBoard/wiki/Splitting-the-configuration
* add info on how to use in the docs.

I'll try, if this works. If it does, I'll leave out the locales from the bundle for the present PR. 

We can then work on the localization separately.

akloeckner: Yes, that works. With the recent commits, the following will work in `config.js` (assuming that we're living off the `build` directory):
```
loadScript('../node_modules/angular-i18n/angular-locale_de-de.js');
loadScript('../node_modules/moment/locale/de.js');
```

I'd now like to bundle the localization files using rollup, as this will allow us to catch any updates. But I have no clue, how this should be done...  :-/

akloeckner: Sorry... I had promised to make i18n a separate PR. :-) So, this one is a very simple now...

akloeckner:
So, this is a simple one now... I'll PR i18n separately...

rchl:
I've briefly looked at using `date-fns` instead, which is preferred in the community now (even https://momentjs.com/ discourages its use), but date-fns doesn't currently support [setting locale globally](https://github.com/date-fns/date-fns/issues/816) so it wouldn't be as flexible. We would need to import all languages we want to support up-front (that seems fine) and then all uses of `date-fns` functions would basically have to be wrapped in a function that uses the chosen locale (which user could define in the config).

I'll wait to hear what you have to say about that before merging this one.

(We would still have to include moment.js for the `chart.js` library)

rchl (approved):

akloeckner:
I think, it's good to stick with moment.js until we sort this out once and for all. Since chart.js uses moments we might even end up deciding to stick with it despite its deprecation.

Wrapping in a function wouldn't really stop me from using it, though. Currently `timeAgo` is the only direct usage of that sort. And it's a wrapper already. (Maybe the functional programming variants would make good wrappers, btw.)

rchl:
BTW. If we advertise the use of something like `moment.locale('de')`, we should rather introduce a config option for it so that we can abstract locale change and potentially replace the underlying implementation in the future.

PR #522, opened 2020-11-09
refactor: move default actions to TILE_DEFAULTS

akloeckner:
also introduce the option to disable actions by setting `action: false` or `secondaryAction: false`
also show "clickable" tick not for fixed tile types, but for tiles with `action` defined

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -482,6 +482,9 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       if ($scope.selectOpened(item)) {
          item._classes.push('-top-entity');
       }
+      if (item.action) {
+         item._classes.push('-clickable');
+      }
 
       return item._classes;
    };
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -80,65 +80,28 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    window.addEventListener('resize', debounce(innerHeightToCSS, 250));
 
    $scope.entityClick = function (page, item, entity) {
-      if (typeof item.action === 'function') {
-         return callFunction(item.action, [item, entity]);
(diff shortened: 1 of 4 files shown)


scripts/controllers/main.js
```diff
@@ -80,66 +80,29 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    window.addEventListener('resize', debounce(innerHeightToCSS, 250));
 
    $scope.entityClick = function (page, item, entity) {
-      if (typeof item.action === 'function') {
-         return callFunction(item.action, [item, entity]);
+      if (item.action === false) {
+         return null;
```
rchl, line 84: I think just `return` would be more correct. Returning `null` for `ng-click` seems like has no special meaning so `undefined` would work just fine also.

scripts/globals/constants.js
```diff
@@ -294,7 +339,63 @@ export const TILE_DEFAULTS = {
...
+      },
+   },
+   [TYPES.LOCK]: {
+      action (item, entity) {
+         return this.$scope.toggleLock(item, entity);
+      },
+   },
+   [TYPES.POPUP]: {
+      action (item, entity) {
+         return this.$scope.openPopup(item, entity);
```
rchl, line 373: This one also took the third argument `item.popup` before.

rchl (changes requested):

scripts/globals/constants.js
akloeckner: Yes. In fact, that was redundant, as the `openPopup` defaults to `item.popup` itself anyway. I dropped when I was trying to find a simpler syntax like
```
action: $scope.openPopup,
```
But I failed to find one. If you can think of one, let me know, please.

Until then, I'll re-introduce the third argument for clarity. Despite its redundancy.

rchl:
Thanks.

PR #524, opened 2020-11-12
feat: i18n

### LINKED ISSUE #327: Simplify locale setting
In order for me to get my correct locale in tileboard I had to add angular-locale for the HEADER_ITEMS (except CUSTOM_HTML) to show the correct locale. 

But for the tiles, and HEADER_ITEMS.CUSTOM_HTML I had to add the moment locale file in order to get the correct locale.

Would it be possible to change so one locale settings will affect both header and tiles?

### LINKED ISSUE #344: localization
How to set global localization to date and time?
now displayed in English
i want to do russian on all pages

akloeckner:
This adds support for internationalization. The following will work in `config.js` (assuming that we're living off the `build` directory):
```
loadScript('../node_modules/angular-i18n/angular-locale_de-de.js');
loadScript('../node_modules/moment/locale/de.js');
```

See also: https://github.com/resoai/TileBoard/pull/521#discussion_r520941850

I'd now like to bundle the localization files using rollup, as this will allow us to catch any updates. But I have no clue, how this should be done... Can you give a hint on that, please? @rchl 

fixes #344
fixes #327

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "angular": "^1.8.2",
     "angular-chart.js": "^1.1.1",
     "angular-hammer": "^2.2.0",
+    "angular-i18n": "^1.8.2",
     "angular-moment": "^1.3.0",
     "angularjs-gauge": "^2.2.0",
     "chart.js": "^2.9.4",
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -92,3 +92,12 @@ export function supportsFeature (feature, entity) {
    return 'supported_features' in entity.attributes
       && (entity.attributes.supported_features & feature) !== 0;
 }
+
+export function loadScript (url) {
+   const xhttp = new XMLHttpRequest();
+   const script = document.createElement('script');
+   xhttp.open('GET', url + '?c=' + Date.now(), false);
+   xhttp.send();
+   script.text = xhttp.responseText;
+   document.head.appendChild(script).parentNode.removeChild(script);
+}
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -3,6 +3,8 @@ import moment from 'moment';
 
 export { moment };
 
+export * from '../i18n';
+
 export const mergeObjects = angular.merge;
 
 export const leadZero = function (num) {
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -1,6 +1,8 @@
 import angular from 'angular';
 import moment from 'moment';
 
+export { moment };
+
 export const mergeObjects = angular.merge;
 
 export const leadZero = function (num) {
--- a/scripts/i18n/de-de.js
+++ b/scripts/i18n/de-de.js
@@ -1,0 +1,5 @@
+import 'angular-i18n/de-de';
+import 'moment/locale/de';
+
+import { extractNgLocale } from './utils.js';
+extractNgLocale();
--- a/scripts/i18n/en-us.js
+++ b/scripts/i18n/en-us.js
@@ -1,0 +1,5 @@
+import 'angular-i18n/en-us';
+import 'moment/locale/en-gb';
+
+import { extractNgLocale } from './utils.js';
+extractNgLocale();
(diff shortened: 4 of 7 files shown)


rchl:
> I'd now like to bundle the localization files using rollup, as this will allow us to catch any updates. But I have no clue, how this should be done... Can you give a hint on that, please? @rchl

That `angular-i18n` has nothing exported, it's just a collection of loose locale files. So we'd need to use one of the rollup plugins to just copy the files from node_modules to the output directory. I can help with that but I wonder which locales are we planning to support?

akloeckner:
I'd suggest:
* ru, pl, de (as developers' locales)
* en, es, fr (as "common" locales)

Or we leave it completely up the user and ask them to copy the locale files to `build` (or whereever) themselves?

rchl:
Actually, there exports too in the smaller individual files.

Those locale files are pretty small so I think it would make sense to just bundle them and force all users to load them. Then the user doesn't have to do the hacky `loadScript` and we can make the experience much easier (just flip a config option).

akloeckner:
If we do bundle them all, it will blow up the built files again. But a selection, yes why not. We'll be bound to get a few request to bundle more of them. But probably not too many in the end...

How would we select one of the angular locales in `config.js`? Moment.js handles it via `moment.locale('de');`.

rchl:
Actually maybe spoke too soon. It might be tricky to bundle multiple locales and dynamically choose which one to load.

For example, you can import a specific lokace like this in app.js:
```js
import angular from 'angular';
import 'angular-i18n/pl-pl';

export const App = angular.module('App', ['hmTouchEvents', 'colorpicker', 'angularjs-gauge', 'angularMoment', 'chart.js']);
```

but I can't think of the top of my head how to make this work dynamically. I'll have to look into that.

> If we do bundle them all, it will blow up the built files again.

Each locale file is 2-4KB so it's not that bad, I would say. Especially if we want to provide just a couple.

> How would we select one of the angular locales in `config.js`

The idea is to just have a `locale` option in the config. But how to implement that, I'm not sure yet (see above :)).

akloeckner:
> The idea is to just have a `locale` option in the config. But how to implement that, I'm not sure yet (see above :)).

A quick research on the web seems to suggest it's going to be just another hacky solution...
https://stackoverflow.com/questions/42788828/angularjs-change-locale-file-on-lang-change
https://stackoverflow.com/questions/31706054/angularjs-dynamic-locale

rchl:
The solution from the second link looks nice and clean. The problem is that those files directly register a `ngLocale` module instead of exporting a plain locale object that we could load dynamically. So I guess it would still take hacks (maybe pre-process those files) to get it done.

akloeckner:
To keep track of the argument here as well: the option should be the way to go, on order to allow swapping underlying libraries. 

https://github.com/resoai/TileBoard/pull/521#issuecomment-731380380

akloeckner:
> The solution from the second link looks nice and clean. The problem is that those files directly register a `ngLocale` module instead of exporting a plain locale object that we could load dynamically. So I guess it would still take hacks (maybe pre-process those files) to get it done.

You triggered my curiousness. Not sure, how hacky you consider the last commit I added to this PR. But maybe it's a starting point into the right direction. It declares a `setLocale` in `utils.js` which is temporarily exported to `config.js` for testing. You can test by adding `setLocale('de-de')` to your `config.js`.

Technically, locales are loaded by importing the `angular` and `moment` locale files. From each `angular` locale, I extract the setup-information, which allows to re-generate the `angular.module()` (and thus overwrite previous versions). There is a module for each locale, because rollup would insist on moving the import statements to the top of the file. Currently, only `de-de` and `en-us` work, but it will be easy to add more locales.

What do you think? Is that a suitable way foward? Or too hacky?

rchl:
> There is a module for each locale, because rollup would insist on moving the import statements to the top of the file.

This is how ES modules work. Rollup is just following the spec. For that reason, I would also not put them after other code as that's just confusing.

After a quick look, it does look very hacky. I think we should be able to do better. Even if it takes to programmatically modify the original locale files so that we can import them the way we want.

akloeckner:
Hmm... OK. It was educational to play around in that depth with angular. One smaller modification would be to mock also `angular.module`. But I believe, I will not go any further with a third approach. The work so far will be archived anyways, so we can pick it up later again...

PR #527, opened 2020-11-16
feat: allow item.classes to be a function

### LINKED ISSUE #518: Feature: Add ability to use anonymous functions in classes property
Feature request.
Hi guys, is it possible to add ability to use anonymous functions in classes property similar as we can use anonymous functions in another properties(icons, state, etc).
It will give us more flexibility to manipulate with css and avoid using complex css selectors. 

example: 

```js
<>
state: function (item, entity) { return entity.attributes.some_attr; }
classes: function (item, entity) { return [list]; }
```

Thanks

akloeckner:
Fixes #518 

We'll also need this, if we want to properly merge `classes` as discussed in https://github.com/resoai/TileBoard/pull/519#discussion_r518407342.


scripts/controllers/main.js
```diff
@@ -470,8 +468,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
          item._classes.push('-th-' + item.type);
       }
 
-      if (item.classes) {
-         item.classes.forEach(function (c) {
+      let item_classes = getItemFieldValue('classes', item, entity);
```
rchl, line 471: Variables should be in `camelCase`.
Maybe also enable this rule: https://eslint.org/docs/rules/camelcase :)

scripts/controllers/main.js
```diff
@@ -470,8 +468,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
...
+      if (item_classes) {
+         if (!Array.isArray(item_classes)) {
+            item_classes = [item_classes];
+         }
+         item_classes.forEach(function (c) {
             item._classes.push(c);
          });
```
rchl, line 478: Good use case for destructuring:

```suggestion
         item._classes.push(...item_classes);
```

scripts/controllers/main.js
akloeckner: :-) Fixed that. But enabling the rule found 19 more places. I don't think i should include those here.

rchl (approved):

PR #530, opened 2020-11-18
feat: support config file selection using a "config" query

anthony2856:
Hello, 

I made this feature to allow several configs based on a queryParams. This way, you can have one config file for living room, another for your bedroom and use the same html files. No need to duplicate all the app.

Just add `?config=bedroom` to the url and the app will load the matching config file "bedroom.js".

What do you think ?


akloeckner:
For the record, here's how it could also be done from a pure user `config.js`: https://github.com/resoai/TileBoard/pull/264#issuecomment-552908653. That's also a proof for a need of this feature.

I am wondering if this feature could be used in malicious ways, if included by default. But since I can't think of one, i might be too cautious.

scripts/index.js
```diff
@@ -6,26 +6,30 @@ import './directives';
...
+      alert(`The "${configName}.js" configuration file has loaded but window.CONFIG is not defined!
+Please make sure that it defines a CONFIG variable with proper configuration.`);
       return;
    }
    // @ts-ignore
    window.window.initApp();
 }
 
+const queryParmas = getJsonFromUrl(document.location.search);
+const configName = queryParmas.config ?? 'config';
```
akloeckner, line 29: Do you really want to allow “falsy“ config files such as `''`?

README.md
```diff
@@ -24,9 +24,10 @@ The project structure has changed in the first versioned (v1.0.0) release. Inste
 
 * Make sure that you have Home Assistant 0.77 or greater installed as only new authentication system is supported from now on
 * Download latest release zip file from https://github.com/resoai/TileBoard/releases and unpack to a directory of your choice
-* In chosen directory rename `config.example.js` to `config.js` and adjust it for your needs
+* In chosen directory rename `config.example.js` to `config.js` or to a specific name (ex: `bedroom.js`) and adjust it for your needs
```
akloeckner, line 27: If this guide is a “getting started“, we should maybe not confuse the user with different configuration files here. Instead we could expand on the issue in the configuration section below.

akloeckner (commented):
I took the liberty and went ahead with a few suggestions...

scripts/dynamic-config.js
```diff
@@ -0,0 +1,11 @@
+export const getJsonFromUrl = (url) => {
```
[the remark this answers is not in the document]
rchl: BTW. Very confusing name for the function.

And custom function shouldn't be needed since there is:
```js
const url = new URL(location.href);
url.searchParams.get('config');
```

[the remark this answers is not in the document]
rchl: > Can we use angular's `$location` directly?

If that's available already then even better.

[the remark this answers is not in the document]
anthony2856: I don't think we can use $location here because the AngularJs app is not bootstrapped at this moment. The application is loaded just after the script. But I'll change with the `new URL()` way ;)
And with the new URL, no need to add another function so no more confusing name 👍

scripts/index.js
anthony2856: It's in the case you don't pass a query param named "config", this case you'll get an empty string

README.md
anthony2856: You're right, i'll change this

anthony2856:
> For the record, here's how it could also be done from a pure user `config.js`: [#264 (comment)](https://github.com/resoai/TileBoard/pull/264#issuecomment-552908653). That's also a proof for a need of this feature.
> 
> I am wondering if this feature could be used in malicious ways, if included by default. But since I can't think of one, i might be too cautious.

I'm not really fan of this way to retrieve several config but it's just my opinion ;) Indeed, this way, even if you pass a config name like `?config=bedroom` you'll probably have an error message saying that the variable `CONFIG` is not defined in `config.js`. But, the error is not in `config.js` but in `bedroom.js`. It's clearely not a huge mistake ;)

anthony2856:
I made some changes, let me know if it's better :)

scripts/index.js
rchl: So then this seems wrong. Nullish coalescing operator will pick the empty string in that case. It should be just `||`, I think.

anthony2856: Yes, but you will have an empty string only if you have `?config=` in you url. An if you don't have the query params, this way you will have `undefined`.
But yes, i can change it to `||` to make it work with both cases

anthony2856: Done ;)

rchl (approved):

akloeckner (approved):
Looks great!

PR #533, opened 2020-11-20
fix(CLIMATE): fix climate state toggle not showing up in hvac mode

### LINKED ISSUE #526: BUG: useHvacMode: true
Hello guys,

When I use "useHvacMode" for my climate entities, I just get a blank title, so it doesn't show "heat" or "off" it's just blank. There is a small spot where I can point and click to show the list with options.
I fixed it using:
`<div class="item-climate--mode" ng-if="(_options = getClimateOptions(item, entity))"
                 ng-click="openSelect(item)">
               <span ng-bind="entity.state"></span>
            </div>`

But this will break it when you are not using the HvacMode.

Also when I use HvacMode my airconditioning works just fine, but my heater when I turn it off I can't turn it back on.

rchl:
Also, show "hvac_action" attribute by default in the subtitle as, when
available, it shows the current state of the climate device which should
be quite useful.

Resolves #526

### DIFF
--- a/scripts/globals/constants.js
+++ b/scripts/globals/constants.js
@@ -303,6 +303,11 @@ export const TILE_DEFAULTS = {
          return this.$scope.openCamera(item, entity);
       },
    },
+   [TYPES.CLIMATE]: {
+      subtitle (item, entity) {
+         return item.useHvacMode ? entity.attributes.hvac_action : undefined;
+      },
+   },
    [TYPES.COVER_TOGGLE]: {
       action (item, entity) {
          return this.$scope.toggleCover(item, entity);
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1228,7 +1228,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.getClimateCurrentOption = function (item, entity) {
-      return item.useHvacMode ? entity.attributes.hvac_mode : entity.attributes.preset_mode;
+      return item.useHvacMode ? entity.state : entity.attributes.preset_mode;
    };
 
    $scope.setClimateOption = function ($event, item, entity, option) {


rchl:
As for the main fix, from looking at HA source code, I believe the `entity.attributes.hvac_mode` attribute never exists.

scripts/globals/constants.js
```diff
@@ -303,6 +303,11 @@ export const TILE_DEFAULTS = {
          return this.$scope.openCamera(item, entity);
       },
    },
+   [TYPES.CLIMATE]: {
+      subtitle (item, entity) {
+         return item.useHvacMode ? entity.attributes.hvac_action : undefined;
```
akloeckner, line 308: We should think about a canonical way to disable features. In the `action` properties we decided to use `false`. I believe, the `state` field also uses `false`? Obviously, `undefined` comes closest to the property not being set at all. Maybe, it's also a difference, if there is a default value? But with the new defaults, we could start to get rid of all the hard-coded defaults over time...

That shouldn't stop you from merging this one, but it might be worth to give it minute of thought...

akloeckner (approved):

scripts/globals/constants.js
rchl: I think that `undefined` makes sense to use for "unset` state at the bottom (defaults) later. Then the next layer, like `entityState` can base its behavior on that and for "unset" do some default behavior while for explicit `false` disable the default behavior.

akloeckner: Makes sense, since the “default behavior“ in `entityState` is indeed something generic for all tile types. :+1:

(I was thinking: why not put the contents of `entityState` into the `state` option? But you're right: that would be a lot of doubling, and the hard-coded way is the most straight-forward way to implement it. At least without a default-default `item`... Good we cleared this!)

Issue #509, opened 2020-10-29
Improve popup timeout / clean up entityClick


akloeckner:
see #496 

We should move the popup timeout variable to be user accessible.

In that yourse we should remove the `page` parameter from `entityClick` as it is not used and it "disturbs" the usual function call context of `(item, entity)`.

PR #535, opened 2020-11-21
fix(popup): popup timeout now accessible to user on $scope

### LINKED ISSUE #496: Popup Timed or Trigger Closure
Is it possble to implement or advise how to use the following features?
- popup card auto close after a certain duration (either since open or since last action?)
- add logic to button on a popup card to close, after performing action.

For example, I have a popup card for music selection, and I would like the popup card to close after you select an option. Right now it will just remain on the popup until you close it or click outside the popup.

akloeckner:
Fixes #496 
See also #509 

This allows users to use the popup timeout feature themselves as outlined in #496 .

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1398,6 +1398,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    $scope.openPopup = function (item, entity, layout) {
       if ($scope.popupTimeout) {
          clearTimeout($scope.popupTimeout);
+         $scope.popupTimeout = null;
       }
       $scope.activePopup = {
          item: item,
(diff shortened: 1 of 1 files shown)


scripts/controllers/main.js
```diff
@@ -1407,8 +1407,8 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
    };
 
    $scope.closePopup = function () {
-      if (popupTimeout) {
-         clearTimeout(popupTimeout);
+      if ($scope.popupTimeout) {
+         clearTimeout($scope.popupTimeout);
       }
```
rchl, line 1412: Set `$scope.popupTimeout` to `null` after clearing the timeout. Also in `openPopup`.

rchl (approved):

PR #537, opened 2020-11-22
fix(CAMERA_STREAM): support non-HLS-capable devices

difelice:
Non HLS compatible devices, like old generation iPads, could try to "stream" video using internal browser capabilities, like Safari in iOS.

Camera Stream not working on legacy iPad:
image

Using `video` tag attributes directly works:
image

Example config:
```js
{
  position: [0, 0],
  width: 3,
  height: 2,
  bgSize: "cover",
  state: false,
  type: TYPES.CAMERA_THUMBNAIL,
  refresh: 5000,
  id: "camera.backyard",
  fullscreen: {
    type: TYPES.CAMERA_STREAM,
    noHls: true,
    objFit: "contain",
    id: "camera.backyard",
    bufferLength: 5,
  },
}
```

### DIFF
--- a/scripts/directives.js
+++ b/scripts/directives.js
@@ -227,22 +227,30 @@ App.directive('cameraStream', function (Api) {
             el.style.height = '100%';
             el.muted = 'muted';
 
-            const len = $scope.item.bufferLength || 5;
+            if (Hls.isSupported()) {
+               const len = $scope.item.bufferLength || 5;
 
-            const config = {
-               maxBufferLength: len,
-               maxMaxBufferLength: len,
-            };
+               const config = {
+                  maxBufferLength: len,
+                  maxMaxBufferLength: len,
+               };
 
-            if (hls) {
-               hls.destroy();
+               if (hls) {
+                  hls.destroy();
+               }
+               hls = new Hls(config);
+               hls.loadSource(url);
+               hls.attachMedia(el);
+               hls.on(Hls.Events.MANIFEST_PARSED, function () {
+                  el.play();
+               });
+            } else {
+               el.src = url;
+               el.setAttribute('playsinline', 'playsinline');
+               el.addEventListener('loadedmetadata', function () {
+                  el.play();
+               });
             }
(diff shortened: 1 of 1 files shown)


rchl:
I feel like HLS script itself should handle that. Either through feature detection or user-agent sniffing. Have you tried raising this as an issue on their repo (https://github.com/video-dev/hls.js)?

rchl:
Actually, in their code example, they are checking support with `Hls.isSupported()`. So it's likely that we should do the same instead of introducing an option.

resoai:
A separate issue with HLS we have is that pausing the stream is not implemented when page is changed or screensaver starts. 

https://github.com/resoai/TileBoard/blob/master/scripts/directives.js#L217

scripts/directives.js
```diff
@@ -227,22 +227,30 @@ App.directive('cameraStream', function (Api) {
...
+                  hls.destroy();
+               }
+               hls = new Hls(config);
+               hls.loadSource(url);
+               hls.attachMedia(el);
+               hls.on(Hls.Events.MANIFEST_PARSED, function () {
+                  el.play();
+               });
+            } else {
+               el.playsInline = 'true';
```
rchl, line 248: It might work (haven't tried) due to how DOM maps properties to dom attributes but there is no such property on HTMLVideoElement (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement).

There is a `playsinline` attribute (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video) so do this instead:
```
el.setAttribute('playsinline', 'playsinline')
```

difelice: I'm not where I saw that, probably on a JSX example. Changed. Thanks.

difelice:
> Actually, in their code example, they are checking support with `Hls.isSupported()`. So it's likely that we should do the same instead of introducing an option.

Added check and works as expected. Thanks
image

rchl (approved):

rchl:
Thanks!

PR #544, opened 2020-11-24
fix(CAMERA_STREAM): pause stream when screensaver active or page changed

### LINKED ISSUE #541: Camera stream not paused when showing screensaver
Reported by @resoai  at https://github.com/resoai/TileBoard/pull/537#issuecomment-732432357

> A separate issue with HLS we have is that pausing the stream is not implemented when page is changed or screensaver starts.
> 
> [https://github.com/resoai/TileBoard/blob/master/scripts/directives.js#L217](https://github.com/resoai/TileBoard/blob/master/scripts/directives.js?rgh-link-date=2020-11-23T21%3A19%3A31Z#L217)

rchl:
Implemented in a way where the playback is stopped when "freezed" (should be "frozen") is `true` and after 5s, if "freezed" doesn't become `false`, the stream is stopped completely.

The stream playing in a popup is only stopped on screensaver activation (as pages can't be changed in that case), while the stream tiles are stopped on screensaver and page change.

Resolves #541

### DIFF
--- a/index.html.ejs
+++ b/index.html.ejs
@@ -76,8 +76,7 @@
             </div>
 
             <div ng-if="activeCamera.fullscreen.type === TYPES.CAMERA_STREAM">
-               <camera-stream item="activeCamera.fullscreen" entity="entity"
-                              freezed="false"></camera-stream>
+               <camera-stream item="activeCamera.fullscreen" entity="entity" freezed="screensaverShown"></camera-stream>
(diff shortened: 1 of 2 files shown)


resoai:
Great! Lets indeed rename `freezed` to `frozen` perhaps?

resoai:
Perfect! Thank you very much!

PR #549, opened 2020-11-26
Addon implementation

anthony2856:
Hello, 

Here is the first implementation of the addon (not using ingress for now, maybe can we implement this later ?)

To get the addon implementation work, you'll have to: 
- create `DOCKER_USERNAME` secret in the repository
- create `DOCKER_PASSWORD` secret in the repository
- change `<DOCKER_USERNAME_HERE>` in `config.json` file with your docker username 

Anthony


.github/workflows/on_release.yaml
```diff
@@ -22,9 +21,25 @@ jobs:
           pushd build
           zip -qr "../TileBoard.zip" .
 
-      - uses: marvinpinto/action-automatic-releases@v1.1.0
+      - uses: svenstaro/upload-release-action@v2
```
rchl, line 24: What's the reason for this change?
I haven't checked thoroughly but I would guess that it's not 100% equivalent (especially in changelog generation)?

addon/CHANGELOG.md
```diff
@@ -0,0 +1,71 @@
+# Changelog
```
rchl, line 1: Is it possible to copy this file on building the docker container and .gitignore those files instead?

addon/README.md
```diff
@@ -0,0 +1,669 @@
+# TileBoard
```
rchl, line 1: Can we copy this file on building container?

.github/workflows/on_release.yaml
anthony2856: I had the following error during `on_release` build and after some google research, i found that the `set-env` was deprecated on November 16th and i found the solution to use `svenstaro/upload-release-action@v2`
Capture d’écran 2020-11-27 à 11 39 31

Maybe, there are other options. But changelog generation isn't done by `yarn release` ?

addon/CHANGELOG.md
anthony2856: No we can't do that :/ Home assistant is looking for those file `CHANGELOG.md` and `README.md` to display readme and changelog for the addon. They are not part of the docker image.

addon/README.md
anthony2856: Same as `CHANGELOG.md`... But maybe different for this one. Maybe the `addon/README.md` should be a Readme for the addon and not a readme for the project. What do you think ?

.github/workflows/on_release.yaml
rchl: The author of `marvinpinto/action-automatic-releases` has already made a fix for that but it's not released yet. So I wouldn't want to switch to another action now since it will likely introduce some different behavior and I don't want to investigate whether switching to it makes sense.

addon/README.md
rchl: I think it would make sense to have a separate readme that is better tailored for the addon and doesn't include information that doesn't apply to the addon.

addon/CHANGELOG.md
rchl: Similarly to `README.md`, I think it would make sense to have addon-specific `CHANGELOG.md` since the addon can include changes that have nothing to do with TIleBoard base code.
Also it wouldn't need to duplicate the changes listed in the root CHANGELOG. It could just say something like: updated to TIleBoard v2.3.4 with a link to the tileboard release.

.github/workflows/on_release.yaml
anthony2856: Ok, we can wait for the release and merge this PR when it will be done ?

addon/CHANGELOG.md
anthony2856: Totally agree. I will made a first version of a changelog for the addon.

addon/README.md
anthony2856: Yes, i will write a new readme for the addon, how to configure, etc...

addon/CHANGELOG.md
anthony2856: After a little reflexion, it's more tricky than i thought... Indeed, every release of the project will make a new version of the addon, but you will have to handle manually the `addon/CHANGELOG.md` file since it's not linked to the project commits...

anthony2856:
I haven't found a solution for now to generate a changelog for the addon. This action has to be manual for now... :/
Still waiting for a new release of `marvinpinto/action-automatic-releases` to merge this. Ok for you ?

rchl:
> Still waiting for a new release of `marvinpinto/action-automatic-releases` to merge this. Ok for you ?

I've just switched to the `latest` tag so it should work already.

You can rebase on master and revert your change to the action and then I'll try it out.

anthony2856:
Done ;)

.github/workflows/on_release.yaml
```diff
@@ -28,3 +27,17 @@ jobs:
           prerelease: false
           files: |
             TileBoard.zip
+  
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
rchl, line 35: I'm gonna create those two:
 - `DOCKERHUB_USERNAME`
 - `DOCKERHUB_TOKEN`

Since using the token is preferred over passowrd.

anthony2856: Ok, i modify this part

rchl: Can you do exactly as I've mentioned (DOCKERHUB_ vs DOCKER_) since I've already created those? Thanks. :)

anthony2856: Oups sorry

anthony2856: Good for me

addon/config.json
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "Tile Board",
+    "version": "2.1.2",
+    "slug": "tile_board",
+    "description": "TileBoard Addon",
+    "url": "https://github.com/resoai/TileBoard",
+    "webui": "http://[HOST]:[PORT:8080]",
+    "startup": "application",
+    "init": false,
+    "image": "<DOCKER_USERNAME_HERE>/tileboard-addon-{arch}",
```
rchl, line 10: I've created a `tileboard` organization (the URL for me is https://hub.docker.com/orgs/tileboard) so I guess this will be just `tileboard`.

anthony2856: Yes, I think. For me it was my username but it was a personnal account..

rchl:
Once the addon is published, will it:
 - appear in the add-on store automatically? I suppose not.
 - will the user have to manually add the repository URL in the HA UI here?
Screenshot 2020-12-01 at 23 24 53

Any other considerations? Is there some documentation for creating the addons somewhere?

anthony2856:
> Once the addon is published, will it:
> 
> * appear in the add-on store automatically? I suppose not.
> * will the user have to manually add the repository URL in the HA UI here?
>   Screenshot 2020-12-01 at 23 24 53
> 
> Any other considerations? Is there some documentation for creating the addons somewhere?

Yes, you'll have to add the git repository manually in hassio

rchl:
I've actually moved the addon to a separate repo at https://github.com/resoai/TileBoard-addon and I'm experimenting with it.

My thoughts and changes so far:
 - updated base build containers from 7.1.0 to 8.0.6
 - changed default port from 8080 to 8090 (8080 is quite popular and easily conflicts)
 - updated naming from "Tile Board" to "TileBoard"
 - I've noticed that the bg images from the images/ directory fail to load. nginx restricting them from being served I guess
 - I think it needs a bit more work to get the user-experience more smooth. At least explain that they should install VSCode or `File Editor` addon to be able to create the config directory and file.
 - Add some more logs to be able to diagnose problems with missing config file more easily
 - I wonder if the addon can automate the setting of the API URL. I guess it might be tricky because we don't know if user wants to connect through IP or some public domain
 - Is HTTPS supported if the user has set up a domain and the certificate (through DuckDns/Let's Encrypt, for example)? If not, is it possible to route traffic through the SSL proxy set up by the Nginx addon?

Those comments are just food for though. Not saying that everything should be handled in here.

anthony2856:
> * I've noticed that the bg images from the images/ directory fail to load. nginx restricting them from being served I gues

No, for this part it's actually my "fault". I was thinking not to provides default images from TileBoard and let the users create it's own `images` folder in `config/www/tileboard` to let him add new images. Due to the symlink i can't link to an existing folder. That's why i remove images folder in Dockerfile.



> * The expected location for the config file could be `/config/$CONFIG_FOLDER/config.js` rather than `/config/www/$CONFIG_FOLDER/config.js` as why should that config be exposed through `/IP/local/config.js`?

Actually there is not need to have this all configs and images in `www` folder you're right. It's possible to change the path in the `rootfs/etc/cont-init.d/tile-board.sh`. Also the `README.md` will need to be updated with this new path.

anthony2856:
Hello @rchl , 
Any news on this addon implementation ?

rchl:
I want to continue working on it but I'm a little short on time at the moment.

Some things that are still to do:
[ ] Figure out a solution for bundling images and allowing the user to add his/her own.
[ ] Handle CI so that release in this repo triggers release for the addon
[ ] Figure out automatic changelog generation and version updating in the addon
[ ] Things I've mentioned in https://github.com/resoai/TileBoard/pull/549#issuecomment-737200454

anthony2856:
> * Figure out a solution for bundling bg images and allowing the user to add his/her own.

This is already possible. Every folder/files present in the configurated folder will be "symlinked" to the TileBoard app. So if the user creates a `images` folder ant puts its images into, they will be linked and could be used in the config file

rchl:
> > * Figure out a solution for bundling bg images and allowing the user to add his/her own.
> 
> This is already possible. Every folder/files present in the configurated folder will be "symlinked" to the TileBoard app. So if the user creates a `images` folder ant puts its images into, they will be linked and could be used in the config file

Yes but then the user needs to also duplicate the bundled images there if he/she wants to use those two.
It's, of course, an acceptable option too but ideally, the bundled images would be accessible regardless if the user provides his own directory.

rchl:
Development is happening at https://github.com/resoai/TileBoard-addon/ and especially with https://github.com/resoai/TileBoard-addon/pull/19, most of my comments will be addressed.

PR #555, opened 2020-11-29
fix(POPUP): merge tile defaults when passing newly created tile object

### LINKED ISSUE #553: Popup when called as secondaryAction are not working properly.
Actions for tiles that defined inside popup that called as `secondaryAction` - are not working.

When i’ve redefined `secondaryAction` for tile and call `this.$scope.openPopup` - tiles inside popup not clickable(action - doesn't work), but if you add lights with (+/-) - actions for those buttons work.

Sample code:

```
{
    type: TYPES.SWITCH,
    subtitle: 'Kitchen',
    position: [1, 3],
    id: 'light.f1_r1_light_base_fan',
    title: 'Test',
    icons: {
       on: 'mdi-spin mdi-fan',
       off: 'mdi-fan',
    },
    states: {
       on: "On",
       off: "Off"
    },
    secondaryAction: function (item, entity) {
       this.$scope.openPopup({
          type: TYPES.POPUP,
          position: [0, 3],
          id: {},
          icon: 'mdi-android',
          title: 'History popup',
          state: false,
          popup: {
             tileSize: 100,
             items: [
                {
                   type: TYPES.SWITCH,
                   subtitle: 'Kitchen1',
                   position: [1, 3],
                   id: 'light.f1_r1_light_base_fan',
                   title: 'Test',
                   icons: {
                      on: 'mdi-spin mdi-fan',
                      off: 'mdi-fan',
                   },
                   states: {
                      on: "On",
                      off: "Off"
                   },
                },
             ]
          }
       }, '');
    },
 }
```

but if you add `action` to second tile `Kitchen1` - it works as expected. 

```
action: function (item, entity) {
    console.log('test');
},
```

looks like #551 not fixed issue completely.

 @rchl  can you please check

rchl:
We've overlooked the fact that user can pass a newly created tile object
to $scope.showPopup() in which case the tile defaults were not merged.

Explicitly merge defaults from the showPopup() call but avoid doing it
multiple times for tiles that have already had defaults merged by
checking a special cache key.

Fixes #553

### DIFF
--- a/scripts/controllers/main-utilities.js
+++ b/scripts/controllers/main-utilities.js
@@ -1,6 +1,8 @@
 import mergeWith from 'lodash.mergewith';
 import { TILE_DEFAULTS, TYPES } from '../globals/constants';
 
+const MERGED_DEFAULTS_KEY = '__merged_defaults';
+
 export function mergeConfigDefaults (pages) {
    for (const page of pages) {
       for (const group of page.groups) {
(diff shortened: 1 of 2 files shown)


alphasixtyfive:
It looks like `openDoorEntry` is actually broken now, although I do remember that it was calling `openPopup` internally. Unfortunately I don't have access to a PC to investigate.

rchl:
When calling it programmatically, you mean?

This test tile appears to work for me:
```js
                  {
                     position: [0, 8],
                     width: 2, height: 2,
                     type: TYPES.DOOR_ENTRY,
                     id: {},
                     icon: 'mdi-phone',
                     title: 'DOOR_ENTRY',
                     state: false,
                     layout: {
                        camera: {
                           type: TYPES.CAMERA,
                           objFit: 'contain',  // https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
                           id: 'camera.weather',
                           refresh: 5000,
                           bgSize: 'contain',
                           fullscreen: {
                              type: TYPES.CAMERA,
                              refresh: 15000, // can be number in milliseconds
                              bgSize: 'contain',
                           },
                        },
                        page: {},
                        tiles: [
                           {
                              position: [0, 0],
                              theme: ITEM_TRANSPARENT,
                              title: 'SCRIPT',
                              id: 'script.test_script',
                              type: TYPES.SCRIPT,
                              state: false,
                              variables: {
                                 value: 'value from script',
                              },
                           },
                        ],
                     },
                  },
```

And if I add:
```
                     secondaryAction(item, entity) {
                        this.$scope.openDoorEntry(item, entity)
                     },
```

then that also opens the popup on long-click.

alphasixtyfive:
I am calling it is a similar way, popup opens up but none of the tiles inside are clickable.

rchl:
I could probably envision some problems if the tiles are generated dynamically but otherwise not sure what might be the issue. I guess we can investigate when you have access to your code.

alphasixtyfive:
Ok, what I have is a `CAMERA` tile which has an `action` assigned to open the door entry:

```
 action: function(i) {
     this.$scope.openDoorEntry(TILES.DOOR_ENTRY_FRONT_GATE, {});
 }
```

For some reason it does look like defaults are not being merged as even `item-clickable` is not present but history does popup on long-press.

rchl:
I guess that's because `TILES` is not part of the initial tiles configuration. We are doing a sweep over registered tiles on load to extend them with default values but that won't extend tiles that are "outside".

Will have to think how to fix that.

alphasixtyfive:
For some reason this also does not work:

```
action: function(item, entity) {
    this.$scope.openDoorEntry({
        type: TYPES.DOOR_ENTRY,
        ...
    });
}
```

So I'm not too sure about `TILES`.

rchl:
Fixing in #664

PR #556, opened 2020-11-29
fix: use polyfill for URLSearchParams

akloeckner:
since it is not available on iOS 9.3.5 (i.e. old iPads e.g.)

see #530, @anthony2856

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
     "chart.js": "^2.9.4",
     "hls.js": "^0.14.16",
     "lodash.mergewith": "^4.6.2",
-    "node-sass": "^5.0.0"
+    "node-sass": "^5.0.0",
+    "url-search-params-polyfill": "^8.1.0"
   },
   "resolutions": {
     "chart.js": "2.9.4"
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -8,6 +8,7 @@ import './controllers/noty';
 import './controllers/screensaver';
 import '../styles/all.less';
 import '@mdi/font/scss/materialdesignicons.scss';
+import 'url-search-params-polyfill';
 
 function onConfigLoadOrError (error) {
    if (error) {
@@ -24,8 +25,7 @@ Please make sure that it defines a CONFIG variable with proper configuration.`);
    window.window.initApp();
 }
 
-const url = new URL(document.location.href);
-const configName = url.searchParams.get('config') || 'config';
+const configName = new URLSearchParams(document.location.search).get('config') || 'config';
 
 const script = document.createElement('script');
 script.src = `./${configName}.js?r=${Date.now()}`;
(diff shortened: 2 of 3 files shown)


rchl:
Can you try this instead: https://github.com/resoai/TileBoard/pull/557 ?

I haven't tried it on a device that doesn't support URLSearchParams but looking at the generated code, it looks like something like a polyfill is added.

Please try first without switching to `URLSearchParams` to check if the URL.searchParams is pollyfilled.
If that doesn't work, please try switching to explicit `URLSearchParams`.

rchl:
And the motivation for doing the way I did it is that it's better to handle this transparently through babel than explicitly through a third-party package.

akloeckner:
I see the advantage. But I can't test it. I get an `undefined` error. I get the same from `master`, actually...

rchl:
Not sure where you getting this `undefined` exactly but maybe try adding this and see if the tablet shows more meaningful error:

```diff
--- a/scripts/globals.js
+++ b/scripts/globals.js
@@ -26,6 +26,8 @@ window.onerror = function (error, file, line, char) {
       'Line: ' + line + ':' + char,
    ].join('<br>');
 
+   alert(text);
+
    Noty.addObject({
       type: Noty.ERROR,
       title: 'JS error',
```

rchl:
I've added an extra commit that fixes notifications not working when there is a syntax error in the config. This should ensure that the user will be able to see where the actual error is.

rchl:
And BTW, I also have an old iPad with iOS 9 and I was able to verify that this fixes the issue. But I'd like you to confirm that also. :)

akloeckner:
It's not on the tablet, but on my desktop Firefox. I get this:
grafik

From master. 2f9cee8 is still ok.

rchl:
With `yarn dev` or `yarn build`?

akloeckner:
yarn build

rchl:
Did you make sure to run `yarn` first to install the new dependency.

akloeckner:
tried now to run `yarn`. it says already up-to-date. Ill runn build again. I'll be back in 3 mins...

akloeckner:
still the same. but since you oviously don't have that... I'll check if there's something offending in my config...

rchl:
Yeah, I don't see that with my config. Neither on Firefox nor Chrome.
You could check with the example config to see if it just loads.
Also, is `yarn dev` working? It should provide a better error message if it also reproduces.

akloeckner:
It's ok with the example configuration. 

With my actual configuration, `yarn dev` also fails. `too much recursion` is what the error seems to be:
grafik

I'm stuck for now... I might find out more in the next days... :-/

akloeckner:
Got it. I have that in my `TILE_DEFAULTS`:
```
   [TYPES.CAMERA]: {
      // Have fullscreen
      fullscreen: {
         type: TYPES.CAMERA,
         //bgSize: 'contain',
      }
   }
```

I'll probably open an issue on it, when I still consider that to be allowable tomorrow morning. :-)

rchl:
Interesting case.

It feels a bit wrong to specify the defaults for two different tiles at once.

But it might make sense for cases when we want to provide different defaults for a given tile, depending on where it's used.

We still have to recursively merge in some cases, like when the user provides a `popup` object for the POPUP tile, for example.

rchl:
I've merged my change after confirming that it's working on iOS 9.

PR #557, opened 2020-11-29
fix: add polyfills for URL/URLSearchParams

rchl:


resoai:
Please let me know what you decide, which one to merge :)

rchl:
I've confirmed that it's working on iOS 9 so will merge.

(BTW. HA login page also seems to work on iOS 9 now so there is no need to use long-lived tokens anymore)

PR #581, opened 2021-01-15
fix: update history tile's start date on state change

airthusiast:
### Proposed change 1

When a history tile is shown 24/7 on a dashboard, new entries get added, but the initial start date never moves forward.
As a result, data piles up, and the initial requested offset increases constantly. After days/weeks, huge amount of data stored in the dataset can cause issues on the client side.

This small change will constantly update the start date and delete any data that is older than (NOW - OFFSET).

Shown time frame will constantly be moving forward.

### Proposed change 2

In some particular cases, especially when entity states get updated rarely (only a few times a day, or even week), the history tile will no longer show any value after a while. (=> moving time window).

Hassio itself has this issue and a workaround is described here: https://community.home-assistant.io/t/add-force-update-support-to-template-sensor/106901

The workaround consists in updating any entity attribute value. This will cause the entity's last_updated value to be updated.

On the history tile, this is easily fixed by adding entity's last_updated value to the watcher.

Watcher will add new data, when state or last_updated changes...

### DIFF
--- a/scripts/controllers/main.js
+++ b/scripts/controllers/main.js
@@ -1455,7 +1455,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       }
 
       const day = 24 * 60 * 60 * 1000;
-      let startDate = new Date(Date.now() - ($scope.itemField('offset', config, entity) || day)).toISOString();
+      const startDate = new Date(Date.now() - ($scope.itemField('offset', config, entity) || day)).toISOString();
 
       Api.getHistory(startDate, entityId)
          .then(function (data) {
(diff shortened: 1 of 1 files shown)


scripts/controllers/main.js
```diff
@@ -1562,29 +1562,50 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
                },
             }, $scope.itemField('options', config, entity));
 
+            // Delete old data
+            const deleteOldData = function () {
+               // Update start date
+               startDate = new Date(Date.now() - ($scope.itemField('offset', config, entity) || day)).toISOString();
```
rchl, line 1568: Why do:
 - new Date()
 - toISOString()
 - and then `Date.parse` on it later

?

Why not have it be a number without all that parsing?

scripts/controllers/main.js
```diff
@@ -1450,7 +1450,7 @@ App.controller('Main', function ($scope, $timeout, $location, Api) {
       }
 
       const day = 24 * 60 * 60 * 1000;
-      const startDate = new Date(Date.now() - ($scope.itemField('offset', config, entity) || day)).toISOString();
+      let startDate = new Date(Date.now() - ($scope.itemField('offset', config, entity) || day)).toISOString();
```
rchl, line 1453: It doesn't look like this variable is used anywhere besides the initial call to `getHistory` so I wouldn't change it to `let` as it gives a false impression that it's reused somewhere later.

In the other place that uses `startDate` variable, you can either shadow this variable with a local one with the same name or just use give your variable a different name.

scripts/controllers/main.js
airthusiast: Yes indeed. Extra parsing removed.

scripts/controllers/main.js
airthusiast: Yep. As startDate was also used in function, followed lint by mistake and changed it. 
Now back as ```const``` again. 👍

rchl (approved):
Looks good. Thanks!

PR #605, opened 2021-01-26
Better compatibility with touch devices (tablets, phones)

romci:
AngularJS's ng-click was having trouble firing consistently on my old iPad with iOS 9.x unless a tile was double-tapped. Replacing the ng-click event handler with Hammer's hm-tap has resolved this and made it a much more enjoyable experience while retaining click handling capabilities for desktop browsers.


rchl:
Unfortunately, there are issues with that approach, see #407.

romci:
Ah, thanks, admittedly in the few hours of testing I did, I've not used any of the components or functionality that exhibit the issue. I'll close the pull request until a better way to resolve this presents itself.

PR #635, opened 2021-02-15
fix(WEATHER): minor fixes for weather icon

rchl:
- Don't override background-repeat property set form '.wu' class
 - Center weather icon in case it's smaller than expected


rchl:
I had this tiny change lying around...

PR #643, opened 2021-02-21
Set locale for date and numbers

mikyjazz:
sources at https://github.com/mikyjazz/TileBoard


sources/angular-locale_de.js
```diff
@@ -0,0 +1,143 @@
+'use strict';
```
rchl, line 1: Those locale files are from `node_modules/angular-i18n` module. There is no need to copy and commit them in the repo. Can be just copied directly from the module on building.

rollup.config.js
```diff
@@ -96,6 +97,7 @@ const config = {
          { files: './favicon.png', dest: `./${outDir}/` },
          { files: './manifest.webmanifest', dest: `./${outDir}/` },
          { files: './images/*.*', dest: `./${outDir}/images/` },
+         { files: './sources/*.*', dest: `./${outDir}/sources/` },
```
rchl, line 100: `sources` is a very generic term. Can we use `locales` instead?

(and as stated in another issue, we can copy directly from node_modules)

rchl (commented):
Good job. We've been exploring this in #524 but got a bit hung up on dynamic loading. Nice that you find some package that abstracts that.

BTW. Can you also add the `locale` option in the `README.md`?

rchl (changes requested):
BTW. Can you also add the `locale` option in the `README.md`?

scripts/app.js
```diff
@@ -1,3 +1,15 @@
...
+      let locale = 'en';
+      if (window.CONFIG && window.CONFIG.locale !== null && window.CONFIG.locale !== undefined) {
+         locale = window.CONFIG.locale.toLowerCase();
+      }
+      // check if available language else default to english
+      const locales = ['it', 'fr', 'es', 'de', 'pt', 'pl', 'nl', 'ru'];
+      if (locales.includes(locale)) {
+         tmhDynamicLocaleProvider.localeLocationPattern('/sources/angular-locale_{{locale}}.js').defaultLocale(locale);
+      }
+   });
```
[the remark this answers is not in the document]
mikyjazz: Hi Rafal,
I put the configuration in app.js because I used the defaultLocale method (leaving the real initialization call to the framework);  "defaultlocale" does not wok in init.js...
I can try to switch to "set" method but it's scope is limited to the controller

sources/angular-locale_de.js
mikyjazz: I know, I was unable in any way to get those files copied from the repository.
Note that all locale files are mutually exclusive so I can't use as many import statements as there are (reason for dynamic loading)
The mere presence in package.json of the angular-i18n package is not sufficient to load the correct files.
This at least in my development environment, I don't know if I'm doing something wrong about it.
So I thought to include the files for the most popular languages but you can obviously include those you want

rollup.config.js
mikyjazz: I voluntarily used a generic term com sources because I was thinking about the possibility of using that folder to insert any other files with dynamic loading for subsequent modifications.
In fact, I have almost ready the modification to also set the room for the WEATHER states

mikyjazz:
I immediately modify the readme (as you will have seen I had anyway modified the config.example file). Let me know what you think about the other three points ...

scripts/app.js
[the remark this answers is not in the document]
rchl: No, the `App.config()` call in `init.js` is exactly equivalent to what you are doing here. So you can just move the code there and it will work the same (you can also drop some of those `window.CONFIG` checks since those are already done there).

sources/angular-locale_de.js
rchl: I can help with that part. It should be doable to do that during build. I don't want to duplicate upstream files like that.

rollup.config.js
rchl: Sorry, but I'd like a standalone locales directory. That will make it clearer to users who want to add new locales.

scripts/app.js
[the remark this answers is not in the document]
mikyjazz: done

rollup.config.js
mikyjazz: done

README.md
```diff
@@ -74,6 +74,11 @@ var CONFIG = {
     * a reconnect will be attempted. If not included in the config file, setting defaults to true.
     */
    pingConnection: true,
+   /* locale: locale used for date and number formats - available locales: it, de, es, fr, pt, ru, nl, pl, en (default).
+    * The system use localization files from the package angular-i18n. 
+    * You can add your locale if it is not included adding the corresponding file from the angular-i18n package into the 'locales' folder.
```
rchl, line 79: Adding custom locales would actually not work because there is a whitelist of locales in the `init.js`. Maybe the whitelist can just be removed? Not sure how loading will behave if an invalid locale is specified.

config.example.js
```diff
@@ -219,19 +218,51 @@ var CONFIG = {
                      //classes: ['-compact'],
                      type: TYPES.WEATHER,
                      id: {},
-                     state: function () {return 'Sunny'}, // https://github.com/resoai/TileBoard/wiki/Anonymous-functions
-                     icon: 'clear-day',
-                     icons: { 'clear-day': 'clear'},
+                     state: function () {return 'Clear, night'}, // https://github.com/resoai/TileBoard/wiki/Anonymous-functions
```
rchl, line 221: Can we keep those weather-related changes separate?

PR #651, opened 2021-02-22
fix(clock): show locale-formatted time

rchl:
This makes clock directive use time formatting from the currently loaded
locale. Also refactored it to be a bit cleaner.

Also fallback to locale date for HEADER.DATE/DATETIME. But since the
example config has formatting explicitly defined, this will probably
have no effect on existing configurations.

### DIFF
--- a/scripts/directives/clock.js
+++ b/scripts/directives/clock.js
@@ -1,51 +1,33 @@
-import { leadZero } from '../globals/utils';
-
 /**
  * @ngInject
  *
  * @type {angular.IDirectiveFactory}
+ * @param {angular.IFilterService} $filter
  */
-export default function ($interval) {
+export default function ($filter) {
    return {
-      restrict: 'AE',
-      replace: true,
-      link: function ($scope, $el, attrs) {
-         const $m = document.createElement('div');
-         const $h = document.createElement('div');
-         const $colon = document.createElement('div');
-         const $postfix = document.createElement('div');
-
-         $m.classList.add('clock--m');
-         $h.classList.add('clock--h');
-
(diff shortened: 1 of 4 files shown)


scripts/directives/clock.js
```diff
@@ -1,52 +1,34 @@
...
-         $colon.textContent = ':';
+      restrict: 'E',
+      template: `
+         <div class="clock--h">{{ hour }}</div
+         ><div class="clock--colon">:</div
+         ><div class="clock--m">{{ minute }}</div
+         ><div class="clock--postfix">{{ postfix }}</div>
+      `,
+      link ($scope, $el, attrs) {
+         const children = $el.children();
```
alphasixtyfive, line 17: I'm wondering if there is a better to address specific children rather than by index. Might be a terrible idea, but how about using `querySelector` picking exact  class name?

rchl: Yeah, I didn't like that too much either.

jQuery would allow to use css selector to find an element (`find('.clock--hr')`) while jqLite that angular includes don't support css selectors. So this is the best way I could find.

rchl:
BTW. When merging, I prefer to use "Squash and merge" as that doesn't introduce unnecessary merge commits and cleans up the branch.

scripts/directives/clock.js
rchl: > I'm wondering if there is a better to address specific children rather than by index. Might be a terrible idea, but how about using `querySelector` picking exact class name?

Not a terrible idea and I can go with that

alphasixtyfive:
Thank you!

PR #662, opened 2021-03-02
feat: GRID groupAlignment option

mikyjazz:
The GRID alignment allows you to divide the groups on several lines; in this case it is necessary 
 to specify the row index on which you want to place the group. Each row will contain all the groups
 with the corresponding index for which the maximum width of the screen will be that of the widest row. 
 With this alignment you will be able to use the screen space to the maximum without having 
 to define groups with a very large height. Group 'row' setting is ignored for HORIZONTALLY and VERTICALLY alignments 
grid_tiles


rchl:
I've pushed some minor formatting/naming changes. Now that that's out of the way I'll focus on the actual feature. But give me some time for that. :)

rchl:
Interestingly my changes don't show up in the diff. Something weird about this branch and all sorts of merges. :)

mikyjazz:
Well done for the naming / style changes ... On our work projects there are the various Jenkins / Travis who do it ... Actually, I also had the question of whether to call the new alignment mode "GRID" or "TABLE "... You choose :)

index.html.ejs
```diff
@@ -247,12 +247,20 @@
...
 
-         <div ng-repeat="group in page.groups track by $index"
-              ng-if="!isHidden(group)" class="group"
-              ng-style="groupStyles(group, page)">
-            <div ng-bind="group.title" class="group-title"></div>
-            <div tile ng-repeat="item in group.items track by $index"></div>
-         </div>
+         <table class="groups-table">
+            <tr ng-repeat="rowIndex in page.rowIndexes">
+               <td class="groups-table-row">
```
rchl, line 252: Do we have to use table/tr/td here? Those elements have some default styles and I'd rather avoid those if possible.

rchl:
In my testing I can see that groups align in the middle vertically while in your screenshot they align to the top. Do you have some custom styles applied?

Screenshot 2021-03-03 at 21 59 45

mikyjazz:
There is a style for table and table row
you must change    vertical-align:middle to vertical-align: top in line 474 of main.less to work
I remember that groups are also used in popup, is this right? If affirmative this must be checked, how I can activate popup?

index.html.ejs
mikyjazz: can you suggest another easy way to do this that also uses the same type of ng-repeat iteration?

rchl: I mean wouldn't it be enough to use `div`s instead with the same markup otherwise?

rchl:
As for testing the popup, I use this tile (you will probably have to change entity IDs to ones that exist in your setup):

```js
                  {
                     position: [4, 0],
                     width: 2, height: 2,
                     type: TYPES.POPUP,
                     icon: 'mdi-android',
                     title: 'POPUP',
                     state: false,
                     id: {},
                     secondaryAction: function (item, entity) {
                        this.$scope.openPopup({
                           type: TYPES.POPUP,
                           position: [0, 3],
                           id: {},
                           icon: 'mdi-android',
                           title: 'History popup',
                           state: false,
                           popup: {
                              tileSize: 100,
                              items: [
                                 {
                                    type: TYPES.SWITCH,
                                    subtitle: 'Office',
                                    position: [1, 3],
                                    id: 'light.everspring_inwall_dimmer_level_3',
                                    title: 'Test',
                                    icons: {
                                       on: 'mdi-spin mdi-fan',
                                       off: 'mdi-fan',
                                    },
                                    states: {
                                       on: 'On',
                                       off: 'Off',
                                    },
                                 },
                              ],
                           },
                        }, '');
                     },
                     popup: {
                        tileSize: 100,
                        items: [
                           {
                              position: [0, 0],
                              type: TYPES.LIGHT,
                              id: 'light.everspring_inwall_dimmer_level_3',
                           },
                        ],
                     },
                  },
```

rchl:
> There is a style for table and table row, you must change vertical-align:middle to vertical-align: top in line 474 of main.less to fix

Sure but I'm wondering what's the good default here and whether it should be changed. Or alternatively, a new option should be provided.

mikyjazz:
I mean I added a style for tables and rows (the 'groups-table' and 'groups-table-row' classes) to use for this functionality ...
About the opportunity not to change the style at line 474, the groups are now rendered in the table rows anyway (which has style vertical-align: middle) ... We can try to understand if something else can be done with a style local in index.html.ejs but then it should be! important and frankly I don't understand why; the only reason for conflict for not making that change I see in the behavior of the popup that I am now going to try ...

index.html.ejs
mikyjazz: with divs you have to deal with alignments, overlaps and all those things that drive a person crazy who tries to align objects ... table does everything right on its own

rchl: Are you sure this matters in this case?

Can you provide a case that would be rendered differently when using table elements and when using divs?

rchl: I've looked at it again and I'm pretty sure that the behavior normally associated with tables is not at play here because you are always creating just **one** table cell (`td`) and putting all groups with the same ID in it. I think the div's would work exactly the same with the current logic.

styles/main.less
```diff
@@ -471,7 +471,7 @@ body {
 .group {
    display: inline-block;
    font-size: 0;
-   vertical-align: middle;
+   vertical-align: top;
```
rchl, line 474: I don't think we should be changing this. It was likely chosen on purpose to achieve a certain style (when using `GROUP_ALIGNS.HORIZONTALLY`). We can choose to override it to `top` for `GRID` since it's a new option. I'm just not certain what would be preferred really...

rchl:
The rows are not aligned properly with `customTheme: CUSTOM_THEMES.WINPHONE` theme for example. I've only tested WINPHONE and HOMEKIT so far so there might be issues in other themes also.

styles/main.less
mikyjazz: with the changes in index.html.ejs the groups are in any case rendered inside the rows of the table that have aligment middle
.groups-table {
    display: inline-block;
    font-size: 0;
    vertical-align: middle;
    margin: 0;
    position: relative;

    & -row {
       padding-top: 35px;
    }
}
referred to the page; the alignment attribute of the group style now refers to the position within the table row and I think it is right "top" - in fact this way it is displayed correctly.
This change has no impact on the popup (which is the only other place where groups are used)

mikyjazz:
i tried to check with all themes ... i also did the comparison with the version before my modification. I found some "strange" behavior in the original version that includes overlapping titles, menu bar; before "fixing everything" I would like to know what the desired / expected behavior is ... It seems strange to me that with a vertical layout the group titles should overlap the tiles or the header or possibly the sidebar ...

mikyjazz:
can you send me a screenshot of this incorrect placement?

rchl:
The issue with the `WINPHONE` theme you should be able to see if you create two groups with the same row index. In that theme they are positioned above/below each other rather than in one row. That's because there is a rule that makes them `display: block` here https://github.com/resoai/TileBoard/blob/7b29fad5d2bff861d226ab6d06eb3c702a4242e3/styles/themes.less#L376-L376

rchl:
> before "fixing everything" I would like to know what the desired / expected behavior is ... It seems strange to me that with a vertical layout the group titles should overlap the tiles or the header or possibly the sidebar ...

I would recommend not to start fixing old issues in this same PR as it can get messy quickly. We can handle them later. Let's just try to get GRID layout to work nicely for now.

When you have a lot of tiles that take a lot of space vertically then the tiles start overlapping header and that's not nice but we should try to find some generic solution for that later since it's not a new problem. For example:

Screenshot 2021-03-04 at 23 15 16

styles/main.less
rchl: There is a behavior change for `GROUP_ALIGNS.HORIZONTALLY` layout relative to the master branch.

See those two screenshots (first from master and second from this branch):

Screenshot 2021-03-04 at 23 21 55
Screenshot 2021-03-04 at 23 21 33


BTW. The formatting of the code you've posed is a bit broken. You should wrap your code blocks with three backticks to have them be rendered nicely. See https://guides.github.com/features/mastering-markdown/

mikyjazz:
Fixed everything ... (What you requested). Since the GRID alignment does not make sense for the MOBILE and WINPHONE themes I have introduced a control that if one of those themes is selected returns the alignment to the default HORIZONTALLY. I've also made sure that the grid is only used for GRID alignment, otherwise the old code is used. I think that's okay ...

index.html.ejs
rchl: What about this issue? Still using table semantics.

index.html.ejs
```diff
@@ -247,11 +247,27 @@
...
+                       ng-if="!isHidden(group) && (group.row || 0) === rowIndex"
+                       class="group-row"
+                       ng-style="groupStyles(group, page)">
+                     <div ng-bind="group.title" class="group-row-title"></div>
+                     <div tile ng-repeat="item in group.items track by $index"></div>
+                  </div>
+               </td>
+            </tr>
+         </table>
+         <div ng-if="!hasGridAlignment()">
```
rchl, line 263: This extra div now breaks the horizontal layout (pushed all context 100% down).
(Also I can see trailing spaces in the div :))

rchl (changes requested):

index.html.ejs
mikyjazz: done

index.html.ejs
rchl: @mikyjazz please have a look at this issue

mikyjazz: it is NOT a issue, anyway changed...

scripts/controllers/main-utilities.js
```diff
@@ -1,8 +1,30 @@
...
 
 const MERGED_DEFAULTS_KEY = '__merged_defaults';
 
+export function calcPageRowIndexes (groupsAlign, pages) {
+   for (const page of pages) {
+      const rowIndexes = [];
+      if (groupsAlign === GROUP_ALIGNS.GRID) {
+         for (const group of page.groups) {
+            const rowIndex = group.row || 0;
+            if (!(rowIndex in rowIndexes)) {
```
rchl, line 12: I'm not sure if this does what was intended. This actually check if the `rowIndexes` array has a value at given `rowIndex` index. Not that it has that specific `rowIndex` value.

For example:
```
1 in [0, 2]
true

2 in [0, 2]
false
```

I think you maybe meant to do `!rowIndexes.includes(rowIndex)`?

styles/main.less
```diff
@@ -486,6 +486,36 @@ body {
    }
 }
 
+.group-row {
+   display: inline-block;
+   font-size: 0;
+   vertical-align: top;
```
rchl, line 492: The whole `group` class is duplicated just to change this one property.

scripts/controllers/main-utilities.js
mikyjazz: damn, yes you're right ... I hate this thing, a statement has different meanings in various languages ...

rchl (approved):
Thanks!

PR #664, opened 2021-03-03
fix(DOOR_ENTRY): non-clickable items when opening popup programmatically

rchl:
The item passed to $scope.openDoorEntry() needs to be merged with
defaults so that the layout items get "action" assigned. This is done
from the $scope.openPopup() a bit later but needs to be done before
creating the popup layout. Merging defaults will be a no-op if defaults
are already merged which will be the case when just clicking a DOOR_ENTRY
tile defined in the config.

Also done the same for other types of popups. That's not strictly
necessary right now but could become necessary in the feature if the
defaults are changed.


alphasixtyfive:
Thank you!

PR #666, opened 2021-03-11
feat(WEATHER): localized number values and support "pressure" key

mikyjazz:
- added filters to localize numeric values according to locale
- changes in CLIMATE tile
- added function filterNumber in $scope to call from anonymous functions (existing function numberFilter does not use locale)

### DIFF
--- a/scripts/directives/headerItem.html
+++ b/scripts/directives/headerItem.html
@@ -23,7 +23,7 @@
       </div>
 
       <div class="header-weather--temperature" ng-if="item.fields.temperature">
-         <span ng-bind="getWeatherField('temperature', item, {})"></span>
+         <span ng-bind="(getWeatherField('temperature', item, {}) | number:1)"></span>
          <span ng-bind="getWeatherField('temperatureUnit', item, {})"></span>
       </div>
 
--- a/scripts/directives/headerItem.html
+++ b/scripts/directives/headerItem.html
@@ -23,8 +23,8 @@
       </div>
 
       <div class="header-weather--temperature" ng-if="item.fields.temperature">
-         <span ng-bind="(getWeatherField('temperature', item, {}) | number:1)"></span>
-         <span ng-bind="getWeatherField('temperatureUnit', item, {})"></span>
+         <span ng-bind="(getWeatherField('temperature', item, {}) | number:1)"></span><!--
+         --><span ng-bind="getWeatherField('temperatureUnit', item, {})"></span>
       </div>
 
       <div class="header-weather--summary" ng-if="item.fields.summary">
--- a/scripts/directives/tile.html
+++ b/scripts/directives/tile.html
@@ -350,31 +350,31 @@
          </div>
 
          <div class="weather-temperature" ng-if="item.fields.temperature">
-            <span ng-bind="(getWeatherField('temperature', item, entity) | number:1)"></span><!--
-            --><span ng-bind="getWeatherField('temperatureUnit', item, entity)"></span>
+            <span ng-bind="(getWeatherField('temperature', item, entity) | number:1)"></span>
+            <span ng-bind="getWeatherField('temperatureUnit', item, entity)"></span>
          </div>
 
          <div class="weather-line -items">
             <span ng-if="item.fields.highTemperature" class="weather-item">
                <i class="mdi mdi-arrow-collapse-up"></i>
-               <span ng-bind="(getWeatherField('highTemperature', item, entity) | number:1)"></span><!--
-               --><span ng-bind=" getWeatherField('highTemperatureUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('highTemperature', item, entity) | number:1)"></span>
+               <span ng-bind=" getWeatherField('highTemperatureUnit', item, entity)"></span>
             </span>
          </div>
 
          <div class="weather-line -items">
             <span ng-if="item.fields.lowTemperature" class="weather-item">
                <i class="mdi mdi-arrow-collapse-down"></i>
-               <span ng-bind="(getWeatherField('lowTemperature', item, entity) | number:1)"></span><!--
-               --><span ng-bind="getWeatherField('lowTemperatureUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('lowTemperature', item, entity) | number:1)"></span>
+               <span ng-bind="getWeatherField('lowTemperatureUnit', item, entity)"></span>
             </span>
          </div>
 
          <div class="weather-line -items" ng-if="effectiveTileSize(page, item).width >= 140">
             <span ng-if="item.fields.humidity" class="weather-item">
                <i class="mdi mdi-water"></i>
-               <span ng-bind="(getWeatherField('humidity', item, entity) | number:0)"></span><!--
-               --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('humidity', item, entity) | number:0)"></span>
+               <span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
             </span>
             <span ng-if="item.fields.windSpeed" class="weather-item">
                <i class="mdi mdi-weather-windy"></i>
@@ -386,8 +386,8 @@
          <div class="weather-line -items" ng-if="effectiveTileSize(page, item).width < 140">
             <span ng-if="item.fields.humidity" class="weather-item">
                <i class="mdi mdi-water"></i>
-               <span ng-bind="(getWeatherField('humidity', item, entity) | number:0)"></span><!--
-               --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('humidity', item, entity) | number:0)"></span>
+               <span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
             </span>
          </div>
          <div class="weather-line -items" ng-if="effectiveTileSize(page, item).width < 140">
@@ -409,8 +409,8 @@
          <!-- DEPRECATED -->
          <div class="weather-line" ng-if="item.fields.apparentTemperature">
             Feels like
-            <span ng-bind="(getWeatherField('apparentTemperature', item, entity) | number:1)"></span><!--
-            --><span ng-bind="getWeatherField('apparentTemperatureUnit', item, entity)"></span>
+            <span ng-bind="(getWeatherField('apparentTemperature', item, entity) | number:1)"></span>
+            <span ng-bind="getWeatherField('apparentTemperatureUnit', item, entity)"></span>
          </div>
 
          <div class="weather-line" ng-if="item.fields.pollen">
@@ -419,8 +419,8 @@
          </div>
 
          <div class="weather-line" ng-if="item.fields.precipProbability">
-            <span ng-bind="(getWeatherField('precipProbability', item, entity) | number:0)"></span><!--
-            --><span ng-bind="getWeatherField('precipProbabilityUnit', item, entity)"></span>
+            <span ng-bind="(getWeatherField('precipProbability', item, entity) | number:0)"></span>
+            <span ng-bind="getWeatherField('precipProbabilityUnit', item, entity)"></span>
             chance of rain
          </div>
          <!-- ///DEPRECATED -->
--- a/scripts/directives/tile.html
+++ b/scripts/directives/tile.html
@@ -350,7 +350,7 @@
          </div>
 
          <div class="weather-temperature" ng-if="item.fields.temperature">
-            <span ng-bind="getWeatherField('temperature', item, entity)"></span>
+            <span ng-bind="(getWeatherField('temperature', item, entity) | number:1)"></span>
             <span ng-bind="getWeatherField('temperatureUnit', item, entity)"></span>
          </div>
 
@@ -365,7 +365,7 @@
          <div class="weather-line -items">
             <span ng-if="item.fields.lowTemperature" class="weather-item">
                <i class="mdi mdi-arrow-collapse-down"></i>
-               <span ng-bind="getWeatherField('lowTemperature', item, entity)"></span><!--
+               <span ng-bind="(getWeatherField('lowTemperature', item, entity) | number:1)"></span><!--
             --><span ng-bind="getWeatherField('lowTemperatureUnit', item, entity)"></span>
             </span>
          </div>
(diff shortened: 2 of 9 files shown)


styles/main.less
```diff
@@ -1685,7 +1685,7 @@ camera_stream {
       margin-right: 10px;
 
       .mdi {
-         margin: 0 0px 0 -3px;
+         margin: 0 0px 0 0px;
```
rchl, line 1688: just `margin: 0` would work but it looks like the whole selector can be removed since default is 0.

mikyjazz: with the value -3 some icons were cut off; of course it can be eliminated altogether

rchl:
I have a test tile that renders like this before (master) and after (this change):

Screenshot 2021-03-11 at 21 37 53

Screenshot 2021-03-11 at 21 38 04

Here is the tile's code:
```js
                  {
                     title: 'Home',
                     position: [6, 6],
                     height: 2.5, width: 2,
                     // classes: ['-compact'], // enable this if you want a little square tile (1x1)
                     type: TYPES.WEATHER,
                     id: 'weather.home',
                     state: function () {return 'Clear, night'}, // https://github.com/resoai/TileBoard/wiki/Anonymous-functions
                     icon: 'clear-night',
                     icons: {
                        'clear-night': 'nt-clear',
                        'cloudy': 'cloudy',
                        'exceptional': 'sunny',
                        'fog': 'fog',
                        'hail': 'sleet',
                        'lightning': 'chancestorms',
                        'lightning-rainy': 'tstorms',
                        'partlycloudy': 'partlycloudy',
                        'pouring': 'rain',
                        'rainy': 'chancerain',
                        "snowy": 'snow',
                        'snowy-rainy': 'sleet',
                        'sunny': 'sunny',
                        'windy': 'hazy',
                        'windy-variant': 'flurries'
                     },
                     states: {
                        "clear-night": "Clear, night",
                        "cloudy": "Cloudy",
                        "exceptional": "Exceptional",
                        "fog": "Fog",
                        "hail": "Hail",
                        "lightning": "Lightning",
                        "lightning-rainy": "Lightning, rainy",
                        "partlycloudy": "Partly cloudy",
                        "pouring": "Pouring",
                        "rainy": "Rainy",
                        "snowy": "Snowy",
                        "snowy-rainy": "Snowy, rainy",
                        "sunny": "Sunny",
                        "windy": "Windy",
                        "windy-variant": "Windy"
                     },
                     fields: { // most of that fields are optional
                        summary: '&sensor.dark_sky_summary.state',
                        temperature: '&sensor.dark_sky_temperature.state',
                        temperatureUnit: '&sensor.dark_sky_temperature.attributes.unit_of_measurement',
                        windSpeed: '&sensor.dark_sky_wind_speed.state',
                        windSpeedUnit: '&sensor.dark_sky_wind_speed.attributes.unit_of_measurement',
                        humidity: '&sensor.dark_sky_humidity.state',
                        humidityUnit: '&sensor.dark_sky_humidity.attributes.unit_of_measurement',

                        list: [
                           // custom line
                           'Feels like '
                              + '&sensor.dark_sky_apparent_temperature.state'
                              + '&sensor.dark_sky_apparent_temperature.attributes.unit_of_measurement',

                           // another custom line
                           'Pressure '
                              + '&sensor.dark_sky_pressure.state'
                              + '&sensor.dark_sky_pressure.attributes.unit_of_measurement',

                           // yet another custom line
                           '&sensor.dark_sky_precip_probability.state'
                              + '&sensor.dark_sky_precip_probability.attributes.unit_of_measurement'
                              + ' chance of rain',
                        ],
                     },
                  },
```

rchl:
It looks like it actually looks correct with your changes. The only thing is that the windSpeed icon still has some negative margin applied.

And of course, all the sensor values are not correct but that is expected as I don't have that sensor in HA.

scripts/directives/tile.html
```diff
@@ -350,69 +350,67 @@
...
-               <span ng-bind="getWeatherField('windSpeedUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('windSpeed', item, entity) | number:0)"></span><!--
+            --><span ng-bind="getWeatherField('windSpeedUnit', item, entity)"></span>
             </span>
          </div>
 
-         <div class="weather-line" ng-repeat="line in item.fields.list">
-            <span ng-bind="getWeatherLine(line, item, entity)"></span>
+         <div class="weather-line -items">
+            <span ng-if="item.fields.windSpeed" class="weather-item">
```
rchl, line 402: Should be `item.fields.presure`

scripts/directives/tile.html
```diff
@@ -350,69 +350,67 @@
...
+               <i class="mdi  mdi-gauge"></i>
+               <span ng-bind="(getWeatherField('pressure', item, entity) | number:0)"></span><!--
+            --><span ng-bind="getWeatherField('pressureUnit', item, entity)"></span>
```
rchl, line 405: I know the code uses `ng-bind` a lot and I don't know why but IMO the `{{ x }}` should be preferred. So this will can for example look much cleaner and have no redundant span elements with:

```
               {{ (getWeatherField('pressure', item, entity) | number:0) }} {{ getWeatherField('pressureUnit', item, entity) }}
```

rchl, line 405: As for spaces between the value and the unit, I think it depends on the unit kind.

It makes sense to have no space before `%` but for `hPa` and `m/s`, for example, I think there should be a space. It both looks better and appears to be preferred when googling those.

rchl (changes requested):

scripts/directives/tile.html
mikyjazz: absolutely yes;  I removed spaces inadvertently, now go to put new spaces :)

mikyjazz: i used ng-bind because everywhere it's done that way ...

scripts/directives/tile.html
mikyjazz: yes...

mikyjazz:
for the pressure it is right to use the appropriate field with the icon I added instead of putting it in the list (also in the example)

scripts/directives/tile.html
mikyjazz: the margins and the spacing of the icons seem ok to me

scripts/directives/tile.html
```diff
@@ -350,66 +350,66 @@
          </div>
 
          <div class="weather-temperature" ng-if="item.fields.temperature">
-            <span ng-bind="(getWeatherField('temperature', item, entity) | number:1)"></span><!--
+            <span ng-bind="(getWeatherField('temperature', item, entity) | number:1)"></span> <!--
```
rchl, line 353: The purpose of those `<!-- -->` comments is to remove space between inline elements. Since you want to have space between those, you can just remove those.

scripts/directives/tile.html
rchl: > i used ng-bind because everywhere it's done that way ...

Yes, I was referring to the TileBoard code in general. I don't see a reason for using `ng-bind` and I'd use mustache syntax in new code myself. But I can clean that up later.

From angularjs documentation (https://docs.angularjs.org/api/ng/directive/ngBind):

```
Typically, you don't use ngBind directly, but instead you use the double curly markup like {{ expression }} which is similar but less verbose.

It is preferable to use ngBind instead of {{ expression }} if a template is momentarily displayed by the browser in its raw state before AngularJS compiles it. Since ngBind is an element attribute, it makes the bindings invisible to the user while the page is loading.

An alternative solution to this problem would be using the ngCloak directive.
```

So it mentions some reason to use ngBind but I think it would be better to just ngCloack instead.

rchl: > the margins and the spacing of the icons seem ok to me

Look at the image I've posted in https://github.com/resoai/TileBoard/pull/666#issuecomment-797035590

The "pressure" icon is cut-off on the left.

scripts/directives/tile.html
mikyjazz: I always put spaces in html using &nbsp;
removing comments

scripts/directives/tile.html
mikyjazz: w
                  title: 'Capurso',
                  row: 0, 
                  height: 2.5, width: 2,
                  //width: 1,
                  //height: 2,
                  items: [
                     {
                        // please read README.md for more information
                        // this is just an example
                        position: [0, 0],
                        height: 2.5, // 1 for compact
                        //classes: ['-compact'],
                        type: TYPES.WEATHER,

have you checked that -3 isn't there yet somewhere?

mikyjazz: about {{}} i can change without problem... I use normally it in python/flask
really don't know angular... this is the first time I meet it ... too many frameworks :( java, pyhon, c#, php.... the head is spinning :) or as we translate in dialect here in Bari, "la capa gira" :) :) :)

scripts/directives/tile.html
rchl: I've cleaned it up.

`&nbsp;` would actually make sense in this case but at the same time the line is not likely to break as it takes so little space.

scripts/directives/tile.html
rchl: The reason it looks good for you, I think is that it triggered the code for the "big" tile rather than the small one. I've removed another margin now and it looks good for me.

rchl: Actually maybe it was not the reason. But it should be fixed now.

rchl (approved):

mikyjazz:
ALWAYS a space must be before units

mikyjazz:
please remove those changes

mikyjazz:
"The International System of Units (SI) prescribes inserting a space between a number and a unit of measurement (being regarded as a multiplication sign) but never between a prefix and a base unit; a space (or a multiplication dot) should also be used between units in compound units."

rchl:
Please find any weather website that puts space before the degrees (celsius) symbol.
I've checked the first two that come to mind and there wasn't any:
 - https://openweathermap.org/
 - https://yr.no

mikyjazz:
not interested in what doing others... SI is law...

rchl:
We're not making a government approved application here. If there is widely applied convention that is not in line with the "law" then we should go for the convention.

I'm gonna look more into it tomorrow but it doesn't seem right what you are suggesting (at least for the degrees symbol).

rchl:
And I'm 100% sure that you don't put space before the percentage sign. Nobody writes `100 %`, or at least shouldn't.

rchl:
I admit that SI says that there should be a space before both the "degrees celsius" symbol and the percentage symbol. It's not a law though but a standard and in practice, that rule is not commonly followed (at least outside of scientific papers). Thus, I'm against having a space for those two.

I know you said you "not interested in what doing others" but the majority (haven't found any exceptions) of weather providers don't put space before those symbols. All of those don't have space: yr.no, openweathermap, the weather channel, google. For example:
Screenshot 2021-03-13 at 11 25 47

The original code wasn't consistent about it so I think it makes sense to make this consistent one way or another. I'm in favor of not having space. If you want to argue against it then I guess we'll have to invite a third-party (the owner of the project) to make the decision.

alphasixtyfive:
While ISO 31-0 dictates that there must be a non-breaking space between value and a symbol, all of the style guides that I've come across are against it. I am always driven by what actually looks and feels right and to me personally an absence of space looks more natural.

mikyjazz:
Very well, the project is yours and you can usually decide what is right or best to do. For my part, coming from the academic world, it is obviously impossible to privilege the aesthetic aspect to the detriment of the rules. Go ahead and change the spacing, I'll continue with my fork ... Just to remind you, I initially corrected the examples where "C" was reported in the temperature unit instead of "°C".
Good job!

scripts/directives/headerItem.html
```diff
@@ -23,7 +23,7 @@
       </div>
 
       <div class="header-weather--temperature" ng-if="item.fields.temperature">
-         <span ng-bind="getWeatherField('temperature', item, {})"></span>
+         <span ng-bind="(getWeatherField('temperature', item, {}) | number:1)"></span>
```
rchl, line 26: Does it actually make sense to show one decimal place for the outside temperature? I would argue that it's not because those numbers are never precise enough for that to matter. And again could give other weather forecasts as an example against it.

rchl: In any way, this is not blocking the review, just wanted to know your opinion.

mikyjazz: This is a sensible question, all the services we normally get information from provide a decimal as precision, then it may or may not make sense to show it in the UI, I like to see it ... It would probably be nice to have the ability to configure this thing with a entry in config.json ...

rchl: OK, I'd prefer not to see it as it would make the brain "ingest" the number more easily without the decimal digits but will ignore it for now.

rchl:
Thanks!

(Note that these changes didn't seem to have anything to do with the CLIMATE tile so I changed the summary accordingly).

mikyjazz:
Yes that's right, there were other changes on climate that I excluded to make the pull request only cover one topic. At this point we have a problem, tell me how we should proceed ... In my fork there will always be the space between numbers and units and the decimal when "I like" ... What do I do? Would I do pull of other changes (but then you'll need to change you those things from you) or our "partnership" ends here?

rchl:
I also have a "custom" branch with my local changes that don't belong upstream (here) because those are either not mature or just too specific. That shouldn't pose much of a problem for creating PRs as long as those PRs don't modify the exact code that you have customized.

You can develop the feature on your custom branch. Once you are ready to submit a PR, you can for example commit it on your custom branch, then create a new fresh branch based on upstream `master` and cherry-pick your change there. Should be a clean cherry-pick in most cases. Then submit that branch for review and address any potential issues on that branch.

Once the PR is merged you can switch back to your custom branch and rebase it on upstream's master. There will likely be conflicts if there were some changes made during code review but should be easy to resolve (just pick the upstream side).

(Alternatively, you could even avoid committing the initial version of your changes on your custom branch to avoid conflicts later, during rebase.)

alphasixtyfive:
> Very well, the project is yours and you can usually decide what is right or best to do. For my part, coming from the academic world, it is obviously impossible to privilege the aesthetic aspect to the detriment of the rules. Go ahead and change the spacing, I'll continue with my fork ... Just to remind you, I initially corrected the examples where "C" was reported in the temperature unit instead of "°C".
> Good job!

I know that you are right and I very much value the time you are spending to push this project forward. I have no scientific or even programming background, I'm an accountant and most of the time whatever I do is dictated solely but a gut feeling.

PR #667, opened 2021-03-13
Localized climate options

mikyjazz:
- localized/customized lookup for climate options
- icon lookup for climate options
- localized translation for mediaplayer "Source" label


config.example.js
```diff
@@ -162,7 +162,23 @@ var CONFIG = {
                      type: TYPES.SENSOR,
                      id: 'updater.updater',
                      state: '@attributes.release_notes' // https://github.com/resoai/TileBoard/wiki/Templates
-                  }
+                  },
+                  {
+                     title: 'Heating',    
+                     position: [1, 1],
+                     id: "climate.000000000",
```
rchl, line 169: Can a value like `{attributes: {current_temperature: 11}` here instead so that there is no popup about missing entity?

scripts/controllers/main.js
```diff
@@ -1288,13 +1320,14 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
 
       let service;
       const serviceData = {};
+      const _option = $scope.reverseLookupClimateOption(item, entity, option);
```
rchl, line 1323: Having both `option` and `_option` variables don't make thing very clear or readable.

In this maybe it can be called `climateMode`.

scripts/directives/tile.html
```diff
@@ -374,43 +375,44 @@
...
                <i class="mdi mdi-water"></i>
                <span ng-bind="(getWeatherField('humidity', item, entity) | number:0)"></span><!--
-               --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
+            --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
             </span>
             <span ng-if="item.fields.windSpeed" class="weather-item">
                <i class="mdi mdi-weather-windy"></i>
-               <span ng-bind="(getWeatherField('windSpeed', item, entity) | number:0)"></span>&nbsp;<!--
-               --><span ng-bind="getWeatherField('windSpeedUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('windSpeed', item, entity) | number:0)"></span><!--
```
rchl, line 382: Should not remove `&nbsp;`

scripts/directives/tile.html
```diff
@@ -374,43 +375,44 @@
...
-               --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
+            --><span ng-bind="getWeatherField('humidityUnit', item, entity)"></span>
             </span>
          </div>
          <div class="weather-line -items" ng-if="effectiveTileSize(page, item).width < 140">
             <span ng-if="item.fields.windSpeed" class="weather-item">
                <i class="mdi mdi-weather-windy"></i>
-               <span ng-bind="(getWeatherField('windSpeed', item, entity) | number:0)"></span>&nbsp;<!--
-               --><span ng-bind="getWeatherField('windSpeedUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('windSpeed', item, entity) | number:0)"></span><!--
```
rchl, line 398: Should not remove `&nbsp;`

scripts/directives/tile.html
```diff
@@ -374,43 +375,44 @@
...
+            --><span ng-bind="getWeatherField('windSpeedUnit', item, entity)"></span>
             </span>
          </div>
 
          <div class="weather-line -items">
             <span ng-if="item.fields.pressure" class="weather-item">
                <i class="mdi  mdi-gauge"></i>
-               <span ng-bind="(getWeatherField('pressure', item, entity) | number:0)"></span>&nbsp;<!--
-               --><span ng-bind="getWeatherField('pressureUnit', item, entity)"></span>
+               <span ng-bind="(getWeatherField('pressure', item, entity) | number:0)"></span><!--
```
rchl, line 406: Should not remove `&nbsp;`

scripts/directives/tile.html
```diff
@@ -577,7 +573,7 @@
                <td colspan="3" class="media-player-table--td-source">
                   <div class="media-player--source" ng-click="openSelect(item)"
                        ng-if="entity.attributes.source_list.length && !item.hideSource">
-                     <span ng-bind="entity.attributes.source || 'Source'"></span>
+                     <span ng-bind="entity.attributes.source || item.textSource || 'Source'"></span>
```
rchl, line 576: Wouldn't it be better if `item.textSource` overrode everything else if set?

But if it's meant as a fallback when the attribute is missing then I'd suggest calling it `sourceFallbackText` instead.

scripts/directives/tile.html
```diff
@@ -495,19 +497,19 @@
             <span ng-bind="climateTarget(item, entity)"></span>
             <span ng-if="(_unit = entityUnit(item, entity))"
                   class="item-climate--target--unit" ng-bind="_unit"></span>
+            <i class="item-climate--icon mdi" ng-class="entityIcon(item, entity)"></i>
```
rchl, line 500: there is a lot of trailing spaces here

scripts/controllers/main.js
```diff
@@ -1274,12 +1261,57 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+      if (typeof item.climateOptions === 'undefined') {
+         const _options = item.useHvacMode ? entity.attributes.hvac_modes : entity.attributes.preset_modes;
+         const options = {};
+         for (const option of _options) {
+            if (item.states !== null && typeof item.states === 'object') {
+               options[option] = item.states[option] || option;
+            } else {
+               options[option] = option;
+            }
+         }
```
rchl, line 1302: I don't have a real entity to test so maybe I'm missing something but I don't understand the point of creating this mapping. It assigns the same value to both the object's key and value so then both the `lookupClimateOption` and `reverseLookupClimateOption` seem to be doing the same thing in practice.

rchl (changes requested):

scripts/controllers/main.js
mikyjazz: really is looked-up option... luOption? :)

scripts/directives/tile.html
mikyjazz: code before your modification in other branch, for.your,eyes,only :)

scripts/directives/tile.html
mikyjazz: idem

scripts/directives/tile.html
mikyjazz: ...

scripts/directives/tile.html
mikyjazz: adjust spaces as you feel... I don't pay attention...

scripts/directives/tile.html
mikyjazz: maybe better change order like:
item.textSource || entity.attributes.source || 'Source'"

config.example.js
mikyjazz: ok

scripts/controllers/main.js
mikyjazz: from my real configuration:
I can have Automatico instead of auto, and so on

                        states: {
                           'auto': 'Automatico',
                           'heat': 'Manuale',
                           'off': 'Spento'
                        },
                        icons: {
                           'auto': 'mdi-calendar-sync',
                           'heat': 'mdi-fire',
                           'off': 'mdi-power'
                        }

mikyjazz: a

mikyjazz: it assign the same value if definition is missing in states...

mikyjazz: to do reverselookup -> settingvalue

mikyjazz: the mechanism is the same with or without states definition...

scripts/directives/tile.html
rchl: Don't understand what you mean.

I'm not talking about indentation change but about removal of `&nbsp;` which has a purpose there.

scripts/controllers/main.js
rchl: Ok, I get know that it has a meaning when user defines states.

scripts/directives/tile.html
mikyjazz: changes in weather are "remaining" of previous changes/pull request; this features was ready at the same time of weather modifications; some changes not tracked/updated rrespect to your latest changes in other pr

mikyjazz: sorry

scripts/directives/tile.html
rchl: sure but you are contributing to a project that cares...

I'll just set up an automatic formatter for HTML files then so we don't have to discuss such things in the future.

scripts/directives/tile.html
rchl: works for me

(we should document that new property also)

rchl: You did document it already. Sorry.

scripts/directives/tile.html
```diff
@@ -495,19 +497,19 @@
...
+            <i class="item-climate--icon mdi" ng-class="entityIcon(item, entity)"></i>                  
          </div>
          <div class="item-climate--mode" ng-if="(_options = getClimateOptions(item, entity))"
               ng-click="openSelect(item)">
             <span ng-bind="getClimateCurrentOption(item, entity)"></span>
          </div>
       </div>
-
       <div ng-if="selectOpened(item)" class="item-select"
            ng-style="itemSelectStyles(entity, _options)">
```
rchl, line 508: The `_options` is an object now so `itemSelectStyles` won't work correctly now.

scripts/controllers/main.js
```diff
@@ -1274,12 +1261,57 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+         item.climateOptions = options;
+      }
+   }
```
rchl, line 1305: Here is my suggestion that reuses `item.states`, merging auto-discovered and user values together if the user did not include some. Also, this will make `item.states` the source of truth instead of creating a separate property for essentially the same thing.

```suggestion
   function initializeClimateStates (item, entity) {
      if (item.__statesExtended === undefined) {
         item.states = item.states || {};
         const options = item.useHvacMode ? entity.attributes.hvac_modes : entity.attributes.preset_modes;
         for (const option of options) {
            if (!(option in item.states)) {
               item.states[option] = option;
            }
         }
         item.__statesExtended = true;
      }
   }
```

Note that I've also renamed the function for clarity.

scripts/controllers/main.js
```diff
@@ -1274,12 +1261,57 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+      initializeClimateOptions(item, entity);
+      for (const [key, value] of Object.entries(item.climateOptions)) {
+         if (value === option) {
+            return key;
+         }
+      }
+      return option;
+   };
+
+   $scope.lookupClimateOption = function (item, entity, option) {
```
rchl, line 1287: I'd suggest name `getClimateOptionText`

scripts/controllers/main.js
```diff
@@ -1274,12 +1261,57 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+      ].join(' - ');
+
+      if (item.filter) {
+         return item.filter(value);
+      }
+
+      return value;
+   };
+
+   $scope.reverseLookupClimateOption = function (item, entity, option) {
```
rchl, line 1277: I'd suggest the name `getClimateOptionFromText` and the last argument could be called `optionText` to differentiate from the `option` which typically refers to the state value.

scripts/controllers/main.js
mikyjazz: ok

scripts/controllers/main.js
mikyjazz: ok

scripts/controllers/main.js
mikyjazz: item.__statesExtended are not only states, they can be also preset_modes that does not belong to entity state... it is an attribute

mikyjazz: so the correct name options...

scripts/directives/tile.html
mikyjazz: itemSelectStyles checks only the existence of parameters, don't use at all

scripts/directives/tile.html
mikyjazz: :D

scripts/directives/tile.html
rchl: It does set marginTop based on the number of items.

scripts/controllers/main.js
rchl: I feel like my suggested naming still works, even if the `preset_modes` are not really entity states. I mean if the user wants to localize `preset_modes` then he has to do that in `item.states` right? So from the user's perspective, it does appear to be "states" still.

(btw. the localization of states could be documented also since it's not entirely obvious how it works, especially with the `hvac_modes` and `preset_modes` being relevant here.

akloeckner:
I'm wondering: Isn't all that logic basically an exerpt of what `entityState` does? 
https://github.com/resoai/TileBoard/blob/4126d1a0e1dee9da5477a6ee03be9fd81bdae7e1/scripts/controllers/main.js#L525

If it is, can you make use of that (even more generic) function?

E.g.:
* `getClimateOptions` => something like mapping `hvacmodes` through `entityState`
* `getClimateCurrentOption` => just map current mode through `entityState`
* `setClimateOption` => that might be the hard case

mikyjazz:
> I'm wondering: Isn't all that logic basically an exerpt of what `entityState` does?
> https://github.com/resoai/TileBoard/blob/4126d1a0e1dee9da5477a6ee03be9fd81bdae7e1/scripts/controllers/main.js#L525
> 
> If it is, can you make use of that (even more generic) function?
> 
> E.g.:
> 
> * `getClimateOptions` => something like mapping `hvacmodes` through `entityState`
> * `getClimateCurrentOption` => just map current mode through `entityState`
> * `setClimateOption` => that might be the hard case

If you look closely at the code you will see that two different properties are mapped
  const option = item.useHvacMode? entity.state: entity.attributes.preset_mode;
I can't imagine how you would like to generalize this climate control specific thing ...

rchl:
@mikyjazz can you update us on what is the status of this on your side?

From my side, I think some of my comments are still unresolved.

mikyjazz:
hi Rafal,
some changes I have implemented in his time while the ones that are missing I have not taken into consideration; if you are interested in the pull request you can make them yourself. For my part, I realized that it is not possible to carry out a version for the public and one for me only so I have decided to carry out only my fork, with the changes I like, the style aspects that I like and without respecting (or at least without asking myself many problems about it) the constraints related to coding style; in the end it will be much easier to integrate in my version what you do in the master one;) See you soon.

rchl:
Merged manually to master. Thanks.

PR #672, opened 2021-03-22
feat(SLIDER): add vertical slider option and improve design

timota:
Add Vertical Slider tile.

Tile doesn't require any input from user, width and height will be calculated automatically based on Tile size. 
Best to use with tile `height` > 1. 
Also, some other options are available for user: `sliderHeight` and `sliderWidth`. If they are defined inside `slider` - they will be used in first place and take precedence via automated calculation. 
Also, Tile supports `icon`. If `icon` doesn't defined it will be taken to consideration in case of automatic calculation. 

Example:
```
{
    position: [0, 0],
    height: 3,
    id: 'light.entity',
    type: TYPES.SLIDER_VERTICAL,
    unit: '%',
    title: 'Slider',
    icon: 'mdi-lightbulb',
    state: false,
    filter: function (value) {
       var num = parseFloat(value) / 2.55 ;
       return num && !isNaN(num) ? num.toFixed() : 0;
    },
    value: '@attributes.brightness',
    slider: {
       max: 255,
       min: 0,
       step: 5,
       field: 'brightness',
       // sliderWidth: '60',
       // sliderHeight: '270',
       request: {
          type: "call_service",
          domain: "light",
          service: "turn_on",
          field: "brightness"
       },
    },
}
```

Screenshot 2021-03-22 at 01 36 06


scripts/controllers/main.js
```diff
@@ -426,6 +426,30 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+      const sliderWidth = item.slider.sliderHeight || (itemWidth - tileSize * width + iconHeight - (item.height === 1 ? 0 : (20 + tileMargin)));
+
+      const styles = {
+         height: sliderHeight + 'px',
+         width: sliderWidth + 'px',
+         top: sliderWidth / 2 - sliderHeight / 2 + 'px',
+         right: sliderHeight / 2 - sliderWidth / 2 + 'px',
+      };
+
+      return styles;
```
rchl, line 450: Thanks for working on this. I have some comments.

Some issues I've observed:
 - the size calculation is not right. The check for icon doesn't work properly when `icon` property is not defined (you are only considering empty string)
Screenshot 2021-03-22 at 22 25 12
 - the styling is not good with that particular theme (`CUSTOM_THEMES.HOMEKIT`). The text is not visible and the slider color blends with the background.
 - if I don't define the `height`, for example, it can end up looking like this:
Screenshot 2021-03-22 at 22 30 23

styles/main.less
```diff
@@ -2243,6 +2243,117 @@ input[type=range]::-moz-range-thumb {
...
+   outline: 0;
+   border: 0;
+   border-radius: @sliderBorderRadius;
+   margin: 0;
+   transition: box-shadow 0.2s ease-in-out;
+   -webkit-transform:rotate(270deg);
+   -moz-transform:rotate(270deg);
+   -o-transform:rotate(270deg);
+   -ms-transform:rotate(270deg);
+   transform:rotate(270deg);
```
rchl, line 2305: Please don't add prefixes for transforms. That can be done on building if not done already.

Also, sort all properties lexicographically.

rchl:
Will also need to be added to TILES.md (example) and README.md (at least the new type)

rchl (changes requested):

styles/main.less
timota: This was fixed. Hope sorting correct.

scripts/controllers/main.js
timota: 2. Added styles for `CUSTOM_THEMES.HOMEKIT`. 
3. height has been fixed. 

1. Formula to count Slider Height is correct. Now i consider case when `item.icon is defined and empty`. 

Hope you will like these changes.

scripts/controllers/main.js
```diff
@@ -426,6 +426,32 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
+      const height = item.height || 1;
+      const tileSize = page.tileSize || CONFIG.tileSize;
+      const tileMargin = page.tileMargin || CONFIG.tileMargin;
+
+      const itemHeight = tileSize * width + tileMargin * (width - 1);
+      const itemWidth = tileSize * height + tileMargin * (height - 1);
+
+      // if item.icon defined, iconHeight should be 0. CSS will count slider height automatically
+      // if item.icon is not defined - we have to count Height.
+      const iconHeight = (item.icon && item.icon !== '') ? 0 : 42;
```
timota, line 440: This will resolve issue when icon defined and empty.

rchl: The `item.icon` condition already covers the empty `item.icon` so just `item.icon` is equivalent to those two conditions.

Also, you are still not handling the case when `icon` is undefined (not set at all).

scripts/controllers/main.js
```diff
@@ -435,10 +435,12 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
       const itemWidth = tileSize * height + tileMargin * (height - 1);
 
-      const iconHeight = item.icon || item.icon === '' ? 0 : 42;
+      // if item.icon defined, iconHeight should be 0. CSS will count slider height automatically
+      // if item.icon is not defined - we have to count Height.
+      const iconHeight = (item.icon && item.icon !== '') ? 0 : 42;
 
       const sliderHeight = item.slider.sliderWidth || itemHeight - 30;
-      const sliderWidth = item.slider.sliderHeight || (itemWidth - tileSize * width + iconHeight - (item.height === 1 ? 0 : (20 + tileMargin)));
+      const sliderWidth = item.slider.sliderHeight || (itemWidth - tileSize * width + iconHeight - (height === 1 ? 0 : (20 + tileMargin)));
```
rchl, line 443: Still have the same problem I've mentioned with slider being rendered outside of the tile if height is not defined.

scripts/directives/tile.html
```diff
@@ -330,7 +330,7 @@
         class="item-entity-container">
 
         <div class="slidervertical-inner">
-         <div class="slidervertical-inner-icon" ng-if="item.icon && item.icon !=''">
+         <div class="slidervertical-inner-icon" ng-if="item.icon && item.icon !==''">
```
rchl, line 333: Again, the `item.icon !==''` condition is redundant.

styles/main.less
```diff
@@ -2287,71 +2287,67 @@ input[type=range]::-moz-range-thumb {
...
-      border-bottom:20px solid @sliderColor;
       -webkit-appearance: none;
-      height: 100%;
-      cursor: ns-resize;
       background: @sliderColor;
-      box-shadow: -350px 0 0 350px @sliderColor, inset 0 0 0 80px @sliderThumbColor;
+      border-bottom:20px solid @sliderColor;
+      border-left:10px solid @sliderColor;
+      border-right:10px solid @sliderColor;
+      border-top:20px solid @sliderColor;
```
rchl, line 2328: Would appreciate having a space after `:` consistently.

styles/themes.less
```diff
@@ -456,6 +456,48 @@
...
+    &::-moz-range-thumb{
+      background: @sliderColor;
+      border-bottom:20px solid @sliderColor;
+      border-left: 12px solid @sliderColor;
+      border-right: 12px solid @sliderColor;
+      border-top:20px solid @sliderColor;
+      box-shadow: -350px 0 0 350px @sliderColor, inset 0 0 0 80px @sliderThumbColor;
+    }
+
+  }
```
rchl, line 499: I would hope there is some more generic way to do that that works in all themes because this is not very scalable.

I would expect to use some theme-specific color variables without repeating all those selectors in each theme.

Also, have you tested that it looks decent in every theme?

styles/main.less
```diff
@@ -2243,6 +2243,113 @@ input[type=range]::-moz-range-thumb {
...
+   }
+
+   &::-moz-range-thumb {
+      background: @sliderColor;
+      border-bottom: 20px solid @sliderColor;
+      border-left: 12px solid @sliderColor;
+      border-right: 12px solid @sliderColor;
+      border-top: 20px solid @sliderColor;
+      border-radius: 0;
+      box-shadow: -350px 0 0 350px @sliderColor, inset 0 0 0 80px @sliderThumbColor;
```
rchl, line 2345: Those box-shadow offsets are rather large. Is that really correct / needed?

scripts/controllers/main.js
```diff
@@ -426,6 +426,32 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
    };
 
+   $scope.sliderVertStyles = function (page, item) {
+      const width = item.width || 1;
+      const height = item.height || 1;
+      const tileSize = page.tileSize || CONFIG.tileSize;
+      const tileMargin = page.tileMargin || CONFIG.tileMargin;
+
+      const itemHeight = tileSize * width + tileMargin * (width - 1);
+      const itemWidth = tileSize * height + tileMargin * (height - 1);
```
rchl, line 436: For performance reasons, this should be cached. Otherwise Angular will run this code a lot.

You can use `cacheInItem` function to calculate this once since it doesn't seem like the result can change at runtime.

rchl (changes requested):

scripts/controllers/main.js
timota: Logic here is that if icon defined set Height to 0, if not defined or empty set it to 42. In this case calculation a bit easier. Probably `iconHeight` is incorrect name for this value, as it works in reverse order.

scripts/controllers/main.js
timota: To be honest i don't have such issue. I've tested in in 2 browsers and both looks good and rendered correctly. 
in a code  `const height = item.height || 1;` which always will have result. 

Can you please shed more light on this?

scripts/directives/tile.html
timota: got you. Removed.

styles/main.less
timota: got you. my fault

scripts/controllers/main.js
rchl: This is the tile I'm testing with:

```js
                  {
                     position: [0, 10],
                     width: 2, /*height: 4,*/
                     type: TYPES.SLIDER_VERTICAL,
                     title: 'SLIDER_VERTICAL',
                     unit: '%',
                     id: 'media_player.kitchen',
                     value: '@attributes.volume_level',
                     slider: {
                        max: 1,
                        min: 0,
                        step: 0.01,
                        request: {
                           type: 'call_service',
                           domain: 'media_player',
                           service: 'volume_set',
                           field: 'volume_level',
                        },
                        field: 'volume_level',
                     }
                  }
```

`TILE_SIZE` is 80 and `TILE_MARGIN` is 3. It looks a bit better when `width` is set to 1 but the slider still doesn't quite fit (even though it's positioned better).

alphasixtyfive:
I'm sorry to interrupt you guys, but here are my two thoughts:

1) This should probably be called `VERTICAL_SLIDER` rather than `SLIDER_VERTICAL`
2) I would think that default theme should not have rounded corners of the slider to better match the overall boxy style yet Homekit theme would be great with it.

scripts/controllers/main.js
rchl: Improving the variable name would definitively help with readability.

scripts/controllers/main.js
timota: i'll take a look.

scripts/controllers/main.js
timota: These props have been added to cache.

styles/main.less
timota: Unfortunately it required. Without it styles are broken

scripts/controllers/main.js
timota: To make it work i have changed a calculation logic a bit. Here an explanation.

```js
      // defines free space used by icon
      const iconSpacer = item.icon ? 0 : 35;

      // if `item.slider.sliderHeight` is defined - set SIZE(height or width) to custom value
      // if `item.slider.sliderHeight` is not defined: make calculation and compare if it greater than (0/20) accordingly .
      // This prevents:
      //    to show very small sliders (less than 20px in this case, probably can be increased)
      //    negative values, which can break alignment
      // If user would like to override automatic calculation appropriate option has to be defined
      const sliderWidth = item.slider.sliderWidth || (x => x > 0 ? x : 0)(itemWidth - 30);
      const sliderHeight = item.slider.sliderHeight || (x => x > 20 ? x : 0)(itemHeight - 100 + iconSpacer);
```

With this - your example works. But didnt test it on tileSize < 80, as i believe, in this case, user has to define custom `sliderHeight/sliderWidth` to make it look nice.

scripts/controllers/main.js
timota: Value was renamed, explanation below.
```js
      // defines free space used by icon
      const iconSpacer = item.icon ? 0 : 35;
```

Hope this makes sense.

styles/themes.less
timota: > I would hope there is some more generic way to do that that works in all themes because this is not very scalable.
I would expect to use some theme-specific color variables without repeating all those selectors in each theme.

This can be changed to use parameterised `mixin`s. 

```less
#lib-range-holder-input() {
  .range-holder-input() {
    generic_attr: attr;
    
    &::-webkit-slider-runnable-track {
      color: @Value1; 
    }
    &::-moz-range-track {
      color: @Value2;
    }
  }
}

@Value1: green;
@Value2: red;

.range-holder input[type="range"] {
    #lib-range-holder-input.range-holder-input();
}

.-theme-homekit {
  
  @Value1: black;
  @Value2: white;
  
  .range-holder input[type="range"] {
    extend_attr: extend;
    #lib-range-holder-input.range-holder-input();
    
    &::-moz-range-track {
      extend_this: this;
    }
  }
}

```

Result:
```css
.range-holder input[type="range"] {
  generic_attr: attr;
}
.range-holder input[type="range"]::-webkit-slider-runnable-track {
  color: green;
}
.range-holder input[type="range"]::-moz-range-track {
  color: red;
}
.-theme-homekit .range-holder input[type="range"] {
  extend_attr: extend;
  generic_attr: attr;
}
.-theme-homekit .range-holder input[type="range"]::-webkit-slider-runnable-track {
  color: black;
}
.-theme-homekit .range-holder input[type="range"]::-moz-range-track {
  color: white;
}
.-theme-homekit .range-holder input[type="range"]::-moz-range-track {
  extend_this: this;
}
```
In this case you can get rid of all those selectors and manage all with variables, but potentially raw css will grow in size as all basic attrs can be populated inside themes selectors. Have to be careful with this.

Another way use parameterised mixins as functions;
```less
#lib-range-holder-input() {
  .range-input(@Val1, @Val2) {
    color: @Val1;
    background: @Val2;
  }
}

@Val1: green;
@Val2: red;

.range-holder input[type="range"] {
  &::-webkit-slider-runnable-track {
    #lib-range-holder-input.range-input(@Val1, @Val2);
  } 
}

.-theme-homekit {
  
  @Value1: black;
  @Value2: white;
  
  .range-holder input[type="range"] {
    &::-webkit-slider-runnable-track {
      #lib-range-holder-input.range-input(@Value1, @Value2);
    }
  }
}
```

result:
```css
.range-holder input[type="range"]::-webkit-slider-runnable-track {
  color: green;
  background: red;
}
.-theme-homekit .range-holder input[type="range"]::-webkit-slider-runnable-track {
  color: black;
  background: white;
}
```
In this case we have to define selectors inside themes as well, but code looks more clear/readable as for me. 

So please choose which you prefer.

timota:
@alphasixtyfive 
1. to be honest i like `VERTICAL_SLIDER` more than `SLIDER_VERTICAL`.  I named it `slider_vertical` as this points to the same tile `family` - slider. If all parties are ok to rename - im ok with this. 

> default theme should not have rounded corners

for me it looks so cubic. at least, probably  `radius:2px` ?

TILE_EXAMPLES.md
```diff
@@ -528,6 +528,42 @@ Similar to sensor, but with an icon.<br>
...
 ```
 
+#### SLIDER_VERTICAL
+Similar to SLIDER (Same rules and calculations are applied. ) but positioned Vertically.<br> 
+Has some extra options: sliderWidth and sliderHeight to define custom Slider width and height.<br>
+![SLIDER](images/tile-screenshots/SLIDER_VERTICAL.png)
+```js
+{
+    position: [0, 0],
+    height: 2,    // Best works with height > 1. In case height=1, better use custom sliderHeight to achieve your needs
```
rchl, line 538: I'm not sure why talking about explicit values like "1". Doesn't everything depend on tile `tileSize` also?

"1" could mean 80 pixels for some but 240 for others.

scripts/controllers/main.js
```diff
@@ -426,6 +426,39 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
       };
    };
 
+   $scope.sliderVertStyles = function (page, item) {
+      // const width = item.width || 1;
+
+      const width = cacheInItem(item, 'width', () => item.width || 1);
```
rchl, line 432: I meant to cache the whole result.

Let me just push the change so that we don't spend too much time doing back-and-forth.

rchl (changes requested):
I've actually come to the realization that I'd like this functionality to be merged into `SLIDER`. It could just have a flag to switch to horizontal mode. Otherwise, everything becomes inconsistent as SLIDER has a generic name but can only be horizontal, and also the styles for both of those are completely different. I think the styles for `SLIDER` don't really work that well for Tileboard as the slider is very small so it would make sense to consolidate that.

I can help with doing that work.

TILE_EXAMPLES.md
timota: Merge tiles is really good idea. I will try to make my best to make it happened. 
Your help more than welcome - many thanks.

timota: And that description has been wrote before i changed the way how size calculated. We can change it last when all features will be implemented. Thanks

rchl: I've pushed some changes to kickstart that.

I'm testing with those 3 tiles:

```js
                  {
                     position: [8, 8],
                     height: 2, width: 2,
                     id: 'media_player.kitchen',
                     value: '@attributes.volume_level',
                     type: TYPES.SLIDER,
                     title: 'SLIDER (legacy)',
                     // subtitle: 'SLIDER (legacy)',
                     legacy: true,
                     // state: false,
                     // bottom: true, // puts slider on bottom
                     slider: {
                        max: 1,
                        min: 0,
                        step: 0.01,
                        request: {
                           type: 'call_service',
                           domain: 'media_player',
                           service: 'volume_set',
                           field: 'volume_level',
                        },
                        field: 'volume_level',
                     },
                  },
                  {
                     position: [10, 8],
                     width: 2, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (new, horizontal)',
                     vertical: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.hallway',
                     value: '@attributes.brightness',
                     filter: function (value) {
                        // Since light brightness range is typically from 0-256,
                        // this filter converts the value to 0-100.
                        var num = parseFloat(value) / 2.55 ;
                        return num && !isNaN(num) ? num.toFixed() : 0;
                     },
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [12, 8],
                     width: 2, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (new, vertical)',
                     vertical: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.hallway',
                     value: '@attributes.brightness',
                     filter: function (value) {
                        // Since light brightness range is typically from 0-256,
                        // this filter converts the value to 0-100.
                        var num = parseFloat(value) / 2.55 ;
                        return num && !isNaN(num) ? num.toFixed() : 0;
                     },
```

timota:
Sorry for a long delay in work, was busy with family/work issues. 

So, This commit contains Vertical and Horizontal sliders. 
Vertical behaves same way as it was before (you can manage width/height on your own or leave it for auto-calculation)

New options:
```
  vertical: true,
  alignRow: false,
```

For Horizontal i have added new option to make Tile look `column` or 'row' aligned:
 - Column looks in similar to old(legacy) one, with some re-arrangements (icon added) and slider positioned horizontally. 
 - Row on other side, re arrange items inside Tile to look in-line started with icon, slider and value. (more on screenshots)

To make it testing easier, below you can find Tiles examples. 

```js
                  {
                     position: [0, 0],
                     width: 2, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (vertical 1)',
                     vertical: true,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 50,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [2, 0],
                     width: 2, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (vertical 2)',
                     vertical: true,
                     alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 30,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [2, 1],
                     width: 1, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (vertical 2)',
                     vertical: true,
                     alignRow: true,
                     unit: '%',
                     icon: '',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 30,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [4, 0],
                     width: 1, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (vertical 3)',
                     vertical: true,
                     // alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [5, 0],
                     width: 0.5, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (vertical 4)',
                     vertical: true,
                     alignRow: true,
                     state: '',
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [0, 2],
                     width: 2, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/col)',
                     vertical: false,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [2, 2],
                     width: 2, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/col)',
                     vertical: false,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [2, 3],
                     width: 1, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/col)',
                     vertical: false,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [4, 2],
                     width: 1, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/col)',
                     vertical: false,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [5, 2],
                     width: 0.5, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/col)',
                     vertical: false,
                     alignRow: false,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [6, 0],
                     width: 2, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [8, 0],
                     width: 2, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [8, 1],
                     width: 1, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     unit: '%',
                     icon: '',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [6, 2],
                     width: 1, height: 2,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [7, 2],
                     width: 3, height: 1,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
                  {
                     position: [7, 3],
                     width: 3, height: 0.5,
                     type: TYPES.SLIDER,
                     title: 'SLIDER (horiz/row)',
                     vertical: false,
                     alignRow: true,
                     state: '',
                     unit: '%',
                     icon: 'mdi-lightbulb',
                     id: 'light.f1_r2_light_tv_wall',
                     value: '@attributes.brightness',
                     slider: {
                        max: 255,
                        min: 0,
                        step: 5,
                        field: 'brightness',
                        // sliderHeight: 120,
                        // sliderWidth: 80,
                        request: {
                           type: "call_service",
                           domain: "light",
                           service: "turn_on",
                           field: "brightness"
                        },
                     }
                  },
```

If you will have questions - please let me know. 


image

alphasixtyfive:
Thanks for great work but I really think that the default theme should stay without rounded corners to keep the style consistent.

timota:
> Thanks for great work but I really think that the default theme should stay without rounded corners to keep the style consistent.

default theme has only `radius 2px` just to smooth edges. On screenshot `homekit` theme.

rchl:
It will take me a while to review this properly so please be patient. :)

Just was quickly checking for now and noticed that it doesn't work that well with smaller tiles (tileSize 80):

Screenshot 2021-04-09 at 22 55 48

rchl:
Actually bigger tile size doesn't really help. It appears it's more about the colors of the input. I think you need to override more of its styles since it's probably picking OS styles (macOS/chrome here).

Screenshot 2021-04-09 at 22 59 59

rchl:
Is the `alignRow` option really needed? It looks rather cramped in the row configuration. And there is a lot of unused space in the tile. What would be the use case for that?

rchl:
Quick update. Looks saner when the slider is not at max value but still not good enough. Not enough contrast to be able to tell where the slider control is.

Screenshot 2021-04-09 at 23 09 15

timota:
> It will take me a while to review this properly so please be patient. :)
> 
> Just was quickly checking for now and noticed that it doesn't work that well with smaller tiles (tileSize 80):
> 
> Screenshot 2021-04-09 at 22 55 48

i will take a look what i can do

timota:
> Actually bigger tile size doesn't really help. It appears it's more about the colors of the input. I think you need to override more of its styles since it's probably picking OS styles (macOS/chrome here).
> 
> Screenshot 2021-04-09 at 22 59 59

Thats really strange. i'm using mac/catalina/chrome and dont have access to WinOS at all. If you can help me to identify colours for that it will be great.

timota:
> Quick update. Looks saner when the slider is not at max value but still not good enough. Not enough contrast to be able to tell where the slider control is.
> 
> Screenshot 2021-04-09 at 23 09 15

same as previous comment - for me colours look nice. I can make them a bit darker to look better. I will try android tablet to and hope i can find golden middle.

rchl:
> Thats really strange. i'm using mac/catalina/chrome and dont have access to WinOS at all. If you can help me to identify colours for that it will be great.

I'm on macOS also, not Windows. I'm on Big Sur though, which can make a difference.

timota:
> Is the `alignRow` option really needed? It looks rather cramped in the row configuration. And there is a lot of unused space in the tile. What would be the use case for that?
> 
> EDIT: But maybe the "cramped" issue is just a matter of adding some more padding. I guess it could be useful for creating tiles that take little space vertically...

 I can be wrong but as for me slider `horizontal/AlignRow` as it should looks like.

scripts/globals/constants.js
```diff
@@ -50,6 +50,7 @@ export const TYPES = {
    CAMERA_STREAM: 'camera_stream',
    SCENE: 'scene',
    SLIDER: 'slider',
+   SLIDER_VERTICAL: 'slider_vertical',
```
timota, line 53: need to remove this as it not required any more

timota:
> > Thats really strange. i'm using mac/catalina/chrome and dont have access to WinOS at all. If you can help me to identify colours for that it will be great.
> 
> I'm on macOS also, not Windows. I'm on Big Sur though, which can make a difference.

can you be so kind and pickup colours that work well on Big Sur ? i can compare them with colours on Catalina and Android.
Thanks

styles/main.less
```diff
@@ -2243,6 +2243,190 @@ input[type=range]::-moz-range-thumb {
...
+//       width: 25px;
+//    }
+// }
+
+// .-horizontal.range-holder input[type="range"] {
+//    &::-webkit-slider-thumb {
+//       border-left: 0.9em solid @sliderColor;
+//       border-right: 0.9em solid @sliderColor;
+//    }
+// }
```
rchl, line 2429: Remove all commented-out code if it's not needed.

rchl (commented):
> can you be so kind and pickup colours that work well on Big Sur ? i can compare them with colours on Catalina and Android.
> Thanks

Actually, I see now what's happening. Your entity is in "off" state so its background is a bit transparent, making the slider stand out. Turn that entity on and you will see the same as me.

rchl:
I haven't checked all themes but it seems to look decent with `border: 1px solid #808080;`

Screenshot 2021-04-09 at 23 55 35

scripts/globals/constants.js
timota: done in upcoming commit

styles/main.less
timota: done in upcoming commit

timota (commented):
these are done

timota:
So... in the latest commit i have rearrange styles a bit to fit into lowest possible(without modification) `tileSize: 90px`.
If `tileSize < 90` - some inputs from user are required:
  - `icon` has to be removed to fit it nicely
  - `state` can be disabled as for me it doesnt make any sense as we have `value`

Below you can find screenshots:

`tileSize: 90px`
Screenshot 2021-04-14 at 13 02 56

`tileSize: 80px` default view. Some tiles are marked `FIXME`
Screenshot 2021-04-14 at 13 05 14

`tileSize: 80px` user modified (removed: icon, state)
Screenshot 2021-04-14 at 13 08 23

Hopefully all these make sense.

timota:
@rchl sorry for bothering, but did you have a chance to review this PR ?

rchl:
Didn't have much time. Will try in the evening.

rchl:
I'll be fixing those as I comment as I don't want to prolong this but just mentioning here so you are up to date.

In the default theme, the text color should be white (to match existing tiles and make the contrast better):

Screenshot 2021-04-19 at 21 09 06

TILE_EXAMPLES.md
```diff
@@ -506,25 +506,39 @@ Similar to sensor, but with an icon.<br>
...
+       field: 'brightness',
+       // sliderWidth: '60',     // Custom slider width
+       // sliderHeight: '270',   // Custom slider height
+       request: {
+          type: "call_service",
+          domain: "light",
+          service: "turn_on",
+          field: "brightness"
+       },
+    },
```
rchl, line 541: A lot needs to be made up to date here. Also update screenshot to use the default theme, for consistency.

scripts/directives/tile.html
```diff
@@ -324,6 +325,36 @@
...
+           class="slider-inner"
+           ng-class="{'slider-inner-direction-column': item.vertical || (!item.vertical && !item.alignRow),'slider-inner-direction-row': !item.vertical && item.alignRow}">
+
+         <div class="slider-inner-icon" ng-if="item.icon">
+            <span class="item-entity--icon mdi" ng-class="entityIcon(item, entity)"></span>
+         </div>
+
+         <div class="slider-inner-value" ng-style="{'order': !item.vertical && item.alignRow ? 1 : 0}">
+            <span ng-bind="getSliderValue(item, entity, _c)"></span>
+            <span ng-if="(_unit = entityUnit(item, entity))" ng-bind="_unit"></span>
```
rchl, line 341: Existing classes need to be reused (with some overrides) instead of creating a completely new one as that is a recipe for creating inconsistencies with existing themes.

rchl:
OK, I did some updates that I wanted. Feel free to try it out and check if I did break something, before merging.

timota (commented):
Except one comment regarding custom slider height - all looks good

scripts/controllers/main.js
```diff
@@ -445,11 +445,15 @@ App.controller('Main', function ($scope, $timeout, $location, Api, tmhDynamicLoc
...
-            const sliderHeight = item.slider.sliderHeight || (x => x > 20 ? x : 0)(itemHeight - 100 + iconSpacer);
+            //  - to show very small sliders
+            //  - negative values, which can break alignment
+            // If the user would like to override automatic calculation, an appropriate option has to be defined.
+            const sliderWidth = item.slider.sliderWidth || (itemWidth - 30);
+            const sliderHeight = item.slider.sliderHeight || (itemHeight - 100 + iconSpacer);
+
+            if (sliderWidth <= 0 || sliderHeight <= 22) {
+               styles.display = 'none';
+            }
```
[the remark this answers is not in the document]
rchl: My thinking was that if the input is not visible then it should not have a layout since then it would be focusable. That said, focusing stuff with the keyboard is probably not something very relevant in TileBoard so it's more a theoretical issue. But OK, will revert that.

timota (commented):
thanks

rchl:
Thanks for your work on this!

PR #676, opened 2021-03-27
fix(WEATHER, WEATHER_LIST): replace darksky with openweathermap in examples

### LINKED ISSUE #558: DarkSky transition
With DarkSky shutting down their API are we able to get MDI icons as the icons in the weather card?

eg. my sensor sensor.bom_mdi_icon_0 shows mdi:weather-sunny

Can I show the icon darksky now shows using that sensor some how?

rchl:
Also update icon mappings.

### DIFF
--- a/TILE_EXAMPLES.md
+++ b/TILE_EXAMPLES.md
@@ -591,12 +591,12 @@ This is a custom tile which can be used for displaying values from different sen
    //classes: ['-compact'], // enable this if you want a little square tile (1x1)
    type: TYPES.WEATHER,
    id: 'group.weather',
-   state: '&sensor.dark_sky_summary.state', // label with weather summary (e.g. Sunny)
+   state: '&weather.openweathermap.state', // label with weather summary (e.g. Sunny)
    // Resolved value must either match one of the supported icons or be mapped
    // to one using the 'icons' option. See the 'icons' option for more information.
-   icon: '&sensor.dark_sky_icon.state',
+   icon: '&weather.openweathermap.state',
    // Use this one if you have an URL of the image to show.
-   //iconImage: '&sensor.dark_sky_icon.state',
+   //iconImage: '&sensor.my_weather_icon.state',
    // A map from sensor's state (key) to icon name (value).
    // The value must match the format:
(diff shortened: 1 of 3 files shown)


LucianTothezan:
Hello. the version here above does not work on my tileboard. i am using openweathermap. the date is not changing if i activate the line for dateformat(please see attachement). also there is an undefined comment in the top right corner. how can i remove it please?
image

rchl:
You can remove `undefind` by adding `state: false` to the tile.

As for the date, are you using the latest version of TileBoard?
And what is `locale` set in your config?

LucianTothezan:
thank You for your reply. plain stupid of me for the state: false issue. that one is solved. as for the locale i am on en-us and using the latest version of tileboard(the update was made by copying the content of the zip file in the existing location of the current tileboard. however in order to keep the config.js content from the old one, i just added the locale line in the config . i did a comparaison for the old and new config.js and i did not notice any others. is there another way of updating tileboard, without losing my custome configuration?)

LucianTothezan:
ok, i finally got it working. i had to overwrite the index.html also. now the question remains on how to make openweathermap to show the forecast for the following days, not for the current day on 1 hour interval? there is a format that worked for me inwhich the forecast was for the next days but it was changing the day and also indexing with one hour on each day that was further in time than the current day(so it was using the same algorithm for +1 hour on the current day, but adding the +1 hour on each day). is there a way to do the forecaat always taking as reference the same hour of the day and not indexing it?

rchl:
I'd recommend starting a thread in https://github.com/resoai/TileBoard/discussions

bosyd:
> ok, i finally got it working. i had to overwrite the index.html also. now the question remains on how to make openweathermap to show the forecast for the following days, not for the current day on 1 hour interval? there is a format that worked for me inwhich the forecast was for the next days but it was changing the day and also indexing with one hour on each day that was further in time than the current day(so it was using the same algorithm for +1 hour on the current day, but adding the +1 hour on each day). is there a way to do the forecaat always taking as reference the same hour of the day and not indexing it?

What Index.html file did you overwrite it with?  I am having the same issue.

PR #677, opened 2021-03-28
feat: support sensors which are timestamps (like uptime sensor)

### LINKED ISSUE #671: Support for timestamps
A few versions ago, HA switched to timestamps for sensors such as `uptime`. I suggest, we add support for that by
* checking for the `device_class` to be `timestamp`
* converting the sensor value using `timeAgo()`
* splitting out the unit to be displayed in the tile.

akloeckner:
* use the default `filter` option for sensors
* parse with `timeAgo` in case we have a timestamp
* include the word `ago`, because timestamps are indeed ago
* add hint how to exclude the word `ago` for the uptime sensor

Localization should be handled separately, and globally for all places where `moment` is used.
In order to allow the user to profit from this further localization, `moment()` should be then be made available by `utils.js`.

closes #671

### DIFF
--- a/scripts/globals/constants.js
+++ b/scripts/globals/constants.js
@@ -448,6 +448,14 @@ export const TILE_DEFAULTS = {
          return this.$scope.callScript(item, entity);
       },
    },
+   [TYPES.SENSOR]: {
+      filter (value, item, entity) {
+         if (entity?.attributes?.device_class === 'timestamp') {
+            return timeAgo(value, false);
+         }
+         return value;
+      },
+   },
    [TYPES.SWITCH]: {
       action (item, entity) {
          return this.$scope.toggleSwitch(item, entity);
--- a/scripts/globals/utils.js
+++ b/scripts/globals/utils.js
@@ -56,9 +56,9 @@ export const playSound = function (sound) {
    audio.play();
 };
 
-export const timeAgo = function (time, omitAgo) {
+export const timeAgo = function (time, withoutSuffix = false) {
    const momentInTime = moment(new Date(time));
-   return momentInTime.fromNow(omitAgo ?? false);
+   return momentInTime.fromNow(withoutSuffix);
 };
 
 export const debounce = function (func, wait, immediate) {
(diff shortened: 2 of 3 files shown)


scripts/globals/utils.js
```diff
@@ -54,9 +54,9 @@ export const playSound = function (sound) {
    audio.play();
 };
 
-export const timeAgo = function (time) {
+export const timeAgo = function (time, omitAgo) {
```
rchl, line 57: Semantically it might be better to provide a default value for the `omitAgo` argument - `omitAgo = false` and then no need for the change below.

Also maybe call the argument the same as moment does - `withoutSuffix` since the current naming is English-specific which could be a bit confusing once we support localizations in moment.

rchl: Also, the date can be in the future and while `momentInTime.fromNow` works with those just fine, the name of the argument is also confusing (to be fair also the name of the method but we probably don't want o change that now).

scripts/globals/constants.js
```diff
@@ -448,6 +448,14 @@ export const TILE_DEFAULTS = {
          return this.$scope.callScript(item, entity);
       },
    },
+   [TYPES.SENSOR]: {
+      filter (value, item, entity) {
+         if (entity?.attributes?.device_class === 'timestamp') {
+            return timeAgo(value, false);
+         }
```
rchl, line 455: Just thinking aloud but I was wondering if `device_class === timestamp` implies that it's the "uptime" type of sensor? Because that change might not make sense for sensors other than uptime types.

But since I don't know the cases where those sensors are used, I won't argue with it and assume that it's a good default.

scripts/globals/utils.js
akloeckner: done.

scripts/globals/constants.js
akloeckner: We didn't want to argue. :-) Still...

I have not found any other standard HA integration with `timestamp` device class. So, uptime seems to be the most common use case. 

However, there are people using custom sensors, such as "last time that file changed". Also for those sensors, `timeAgo` is a decent default for a dashboard display.

Then, there might be power users that really want to see the actual timestamp. They would have to change the `filter` then.

rchl (approved):

PR #706, opened 2021-04-27
refactor(CLIMATE): create separate directive

rchl:
Experimenting with separating each tile's code into its own directive.

To summarize, the specific tile's code is separate into its own directive and HTML template.

The directive's controller contains tile-specific code and is exposed as `ctrl` in the template.

Since `item` and `entity` are bound to the isolated scope of the directive, there is no need to pass those from the template anymore since the controller has access to them already. Only in case, we want to call the APIs from the main controller we need to pass them. That part is a bit unclean right now but could be made better eventually.

Please don't merge yet. Still figuring out if this is the best way to do it.

Please have a look and provide any feedback @alphasixtyfive @akloeckner

### DIFF
--- a/scripts/directives/tile.html
+++ b/scripts/directives/tile.html
@@ -518,49 +518,7 @@
       </div>
    </div>
 
-   <div ng-if="item.type === TYPES.CLIMATE" class="item-entity-container">
-      <div>
-         <div
-            class="item-button -center-right"
-            ng-if="entity.attributes.temperature && entity.state !== 'off'"
-            ng-click="increaseClimateTemp($event, item, entity)"
-         >
-            <i class="mdi mdi-plus"></i>
-         </div>
-         <div
-            class="item-button -bottom-right"
-            ng-if="entity.attributes.temperature && entity.state !== 'off'"
-            ng-click="decreaseClimateTemp($event, item, entity)"
-         >
-            <i class="mdi mdi-minus"></i>
-         </div>
-      </div>
-
-      <div class="item-climate">
-         <div class="item-climate--target">
-            <span ng-bind="climateTarget(item, entity)"></span>
-            <span ng-if="(_unit = entityUnit(item, entity))" class="item-climate--target--unit" ng-bind="_unit"></span>
-            <i class="item-climate--icon mdi" ng-class="entityIcon(item, entity)"></i>
-         </div>
-         <div
-            class="item-climate--mode"
-            ng-if="(_options = getClimateOptions(item, entity))"
-            ng-click="openSelect(item)"
-         >
-            <span ng-bind="getClimateCurrentOption(item, entity)"></span>
-         </div>
-      </div>
-      <div ng-if="selectOpened(item)" class="item-select" ng-style="itemSelectStyles(entity, Object.keys(_options))">
-         <div
-            class="item-select--option"
-            ng-repeat="option in _options track by $index"
-            ng-class="{'-active': option === lookupClimateOption(item, entity, entity.state)}"
-            ng-click="setClimateOption($event, item, entity, option)"
-         >
-            <span ng-bind="option"></span>
-         </div>
-      </div>
-   </div>
+   <div ng-if="item.type === TYPES.CLIMATE" class="item-entity-container" tile-climate item="item" entity="entity"></div>
 
    <div ng-if="item.type === TYPES.MEDIA_PLAYER" class="item-entity-container">
       <div class="media-player-table" ng-class="{'-has-state': _state, '-has-subtitle': _subtitle}">
--- a/scripts/directives/tile.js
+++ b/scripts/directives/tile.js
@@ -9,7 +9,7 @@ export default function () {
    return {
       restrict: 'AE',
       replace: false,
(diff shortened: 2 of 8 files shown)


scripts/directives/tile.js
```diff
@@ -9,7 +9,7 @@ export default function () {
    return {
       restrict: 'AE',
       replace: false,
-      scope: true,
+      scope: false,
```
rchl, line 12: Not sure if we needed to create extra scope for each tile. Maybe we did since we are doing stuff like `ng-if="_foo = fn()"` and that avoided possible overwrites. Will have to test how this works exactly.

scripts/directives/tile.html
```diff
@@ -518,49 +518,13 @@
...
-            ng-repeat="option in _options track by $index"
-            ng-class="{'-active': option === lookupClimateOption(item, entity, entity.state)}"
-            ng-click="setClimateOption($event, item, entity, option)"
-         >
-            <span ng-bind="option"></span>
-         </div>
-      </div>
-   </div>
+   <div
+      ng-if="item.type === TYPES.CLIMATE"
```
akloeckner, line 522: Can't we get rid of those specialized code sections completely in `tile.html`? Kind of `include all tile templates from directory ...`?

scripts/directives/tiles/index.js
```diff
@@ -0,0 +1,4 @@
+import { App } from '../../app';
+import tileClimate from './tileClimate';
```
akloeckner, line 2: Also here: Can we generically include all tile types? Alternatively: We could add the `html` part of the template here, so we don't have two places to add a new tile type...

scripts/directives/tile.js
akloeckner: Will this influence, if generic function from the `main` controller are available? I'm thinking of things such as `cacheInItem` or the function you aliased in "Temporary transitioning APIs".

akloeckner (commented):
I am by no means an expert in telling you how AngularJS should be set up. So, I can only give comments on the generic structure. Overall: I think, it's a good idea to separate the specialized tile code from the overall tileboard code. :+1: We should just make sure not to set up any "traps" in doing this. :-)

akloeckner:
> Please don't merge yet. Still figuring out if this is the best way to do it.

Isn't there a "draft PR" feature for this situation?

scripts/directives/tile.html
rchl: Yes, I also thought that would look weird once all tiles are converted to have these repetitive elements here.

I think it could be solved with some directive that would dynamically insert a relevant directive based on the type. My angularjs is a bit rusty but it should be doable, I think.

scripts/directives/tiles/index.js
rchl: Importing all tile files at once might not be doable with ES imports since imports are hoisted and run before any code so we can't read files from disk and then import based on that. But we could do some preprocessing with rollup to generate files with everything imported, maybe.

scripts/directives/tile.js
rchl: `scope: true` creates a new scope that inherits from the parent scope. So everything from the parent scope is still accessible. It can only affect things when setting properties (like in the case I've mentioned).

scripts/directives/tiles/index.js
akloeckner: Maybe it's of lesser importance and we should solve the html part first. (Maybe with said automatic/dynamic directives you mentioned.)

scripts/directives/tile.html
alphasixtyfive: This is good idea. This would also pave the way to completely custom tile types user can create themselves and share if they want.

scripts/directives/tile.js
rchl: And `scope: false` of course runs in the same scope so everything is accessible.

cgarwood:
Any more movement in this direction? I've been wanting to add some new tiles but the current structure makes that a bit difficult, so having them split out to separate files would be awesome.

rchl:
Sadly, I don't think I can currently devote that much time to get all of those refactors done.

Though I guess this PR might be "mergable" already, if we ignore the raised improvement suggestions. Maybe that would kick-start more changes from other people.

cgarwood:
I'd certainly be up for helping split some more tiles out to separate files if there was a framework in place for it 🙂

akloeckner:
My 50ct: Yes, please ignore my suggestions, if they are blocking improvement! If there is interest in advancing this, we should go with the usual spirit... Improve step by step and handle problems if and when they arise. 

So, Charles, if you are planning on adding some tiles, you will certainly stumble accross some of these issues. Meaning, they'd even be adressed in a timely manner. And by the time you're done with the new tiles, the framework will have settled to something consistent. I'd be happy to provide some more 50ct from my non-ng-expert perspective any time, if you like.

If we don't want this convergence to happen on the main branch, we could set up a feature branch and merge when appropriate. However, since the main branch receives updates only once per month by now, I would just wait for this convergence to settle before creating a new release...

PR #724, opened 2021-05-22
fix(HISTORY): fix tooltip time for some locales

akloeckner:
Apparently, the new localization features convert the `label` of the `tooltipItem` to something local.

This results in something like `Mai 23, 2021, 1:03:51`, for me in German. If we try to convert that using `timeAgo`, it fails.

The fix uses the raw `dataset[].data[].x`, which is `Sun May 23 2021 01:03:51 GMT+0200 (Mitteleuropäische Sommerzeit)` in my case. This value can be converted to a time ago properly.

### DIFF
--- a/scripts/globals/constants.js
+++ b/scripts/globals/constants.js
@@ -221,7 +221,7 @@ export const MINIMAL_CHART_OPTIONS = {
          display: true,
          ticks: {
             mirror: true,
-            callback: function (value, index, values) {
+            callback (value, index, values) {
                if (index === values.length - 1 || index === 0) {
                   return value;
                }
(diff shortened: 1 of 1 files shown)


rchl:
Why do you think that this is universally a good change? The localized seems much more readable than the raw value.

I understand the issue with not being able to use `timeAgo` but I don't think this warrants making it worse for the common case. Maybe there is another solution to that?

akloeckner:
I was not clear enough. The fix uses the raw value *and* converts it to `timeAgo`.

The current code is actually broken for me. It shows `invalid date` in the tooltip. So, this is actually really only a fix.

rchl:
Thanks. Just did a minor style change.

PR #737, opened 2021-07-20
Added an example for the VACUUM tile, including a screenshot

sbrandsborg:
Added an example for the VACUUM tile, including a screenshot


alphasixtyfive:
Thanks!

PR #746, opened 2021-08-05
fix(light): fix light brightness adjustment buttons after HA update

### LINKED ISSUE #743: No +/- in TYPES.LIGHT
I use last versions of TileBoard with Home Assitant and zWaveJS.
But I cannot control the light intensity because the + and - are not shown into the tile.

HUE lights (still) get these controls, but somehow I don't get it with my zwave devices.
Anyone else having this problem? Or is it a bug?

(Screenshot are both dimmers, Hal = Hue, Keuken = zwave)

image

rchl:
Take into account the new "supported_color_modes" attribute when
checking if light supports brightness adjustment.

Fixes #743

### DIFF
--- a/scripts/directives/tile.html
+++ b/scripts/directives/tile.html
@@ -142,19 +142,11 @@
 
    <div ng-if="item.type === TYPES.LIGHT" class="item-entity-container">
       <div ng-if="!item.controlsEnabled">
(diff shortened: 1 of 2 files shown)


alphasixtyfive:
Thanks!

PR #793, opened 2021-11-17
feat: create docker images for standalone use

chetan:
Hi,

I noticed there's no official docker build for the project. The below is a simple build which serves using `nginx:alpine`. If you'd like to set up an official repository, let me know and I can update the repository name in the readme below.

chetan

### DIFF
--- a/.github/workflows/on_push.yaml
+++ b/.github/workflows/on_push.yaml
@@ -13,7 +13,6 @@ jobs:
     name: Lint and Build
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout code
         uses: actions/checkout@v2
 
@@ -25,3 +24,22 @@ jobs:
 
       - name: Build
         run: yarn build
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Docker build
+        run: |
+          docker buildx build \
+            --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7 \
+            --pull \
+            -t tileboard/tileboard:latest \
+            -f docker/Dockerfile.run \
+            --push \
+            .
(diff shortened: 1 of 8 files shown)


rchl:
> If you'd like to set up an official repository, let me know and I can update the repository name in the readme below.

I'm OK with having an official docker image for this project but then I'd expect the docker images to be made automatically from CI. Similar to how it's done from the `tileboard-addon` repo - https://github.com/resoai/TileBoard-addon/blob/b6a67d80c82edb634503110b31de91ed6e1bbb79/.github/workflows/on_tag.yaml#L22-L22

chetan:
@rchl added the workflow step for building and pushing the docker image. Have not tested all the way through, but it should work.

.github/workflows/on_push.yaml
```diff
@@ -25,3 +24,22 @@ jobs:
 
       - name: Build
         run: yarn build
+
+      - name: Login to DockerHub
```
rchl, line 28: I think I would prefer to have images created only for tagged releases (from `on-tag.yaml`).

rchl (changes requested):

docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:latest AS builder
```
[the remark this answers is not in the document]
chetan: This image is only used for building and the larger image is usually safer in case anything needs to be compiled for one of the various dependencies in the chain. Pinning to a version is a good idea though.

.github/workflows/on_push.yaml
chetan: Could have a `:latest` (or `:git`) docker tag that tracks the main branch in addition to the specific tags. Totally up to you, though.

.github/workflows/on_push.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           docker buildx build \
             --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7 \
             --pull \
-            -t tileboard/tileboard:latest \
+            -t tileboard/tileboard:master \
```
rchl, line 42: Since we are building tagged releases now, `latest` tag is more appropriate. `master` would be built from every commit.

rchl (approved):

rchl:
Thanks.

PR #796, opened 2021-11-25
feat: allow custom classes on screensaver slides

rsmhass:


rchl:
Should be also mentioned in the readme. Otherwise it's a feature that nobody will know about.

rsmhass:
> Should be also mentioned in the readme. Otherwise it's a feature that nobody will know about.

I was thinking of updating https://github.com/resoai/TileBoard/wiki/Screensaver-configuration rather than updating the README.md. README has lesser of screensaver details than the wiki page. If I update the README with just this feature, then that would be off.

I am assuming I would have the permission to edit https://github.com/resoai/TileBoard/wiki/Screensaver-configuration

rchl:
Yes, readme should regardless list all supported options, even if it does not go into details.

rsmjs:
@rchl I have added documentation for the feature to README. Can this be merged now ?

rsmhass:
@rchl I have added documentation for the feature to README. Can this be merged now ?

PR #823, opened 2022-03-27
Add a tip to Custom CSS Styles section

jiayuandeng:


rchl:
Updated slightly as the text was supposed to be bold and not a list item, if you check other places that have similar tips.

Thanks!

PR #822, opened 2022-03-10
feat: make tileboard installable as web apk

akloeckner:
fixes #819

This
* adds a dummy service worker with a fetch handler
* increases the favicon size to 512px and adds a vector source graphic
* adds a start_url to the webmanifest

The service worker's scope is automatically just the path to its js file. (See [Google's primer on it](https://developers.google.com/web/fundamentals/primers/service-workers#register_a_service_worker).)

The ingress mode of the addon can disable the service worker via `window.IN_INGRESS = true`. See PR there: https://github.com/resoai/TileBoard-addon/pull/32.

Issue #819, opened 2022-02-28
Samsung phones: Progressive Web App lacks service worker


akloeckner:
I wondered why TileBoard wouldn't install as a PWA on my new samsung phone. It appears,  we would need a (at least) dummy service worker installed. 

Here's a list of requirements: https://superpwa.com/doc/test-pwa-samsung-internet-browser/

rchl:
Are you sure this information are up-to-date? Chrome didn't seem to require a service worker and Samsung browser is based on Chromium. Of course they could have changed that condition but not sure why would they.

Can you check on Chrome browser if you can install it?

akloeckner:
It appears, there is a difference between those two constructs:
* `add to home screen` adds a shortcut to the home screen, i.e, the pages that hold such shortlinks or widgets. It will open the TileBoard page in full screen mode as intended. But it cannot be moved to the "app screen", which is "hidden" by a gesture behind the home screen. So, this is basically a widget-like thing and even displays with a small Chrome or SamsungInternet icon. This shortcut can be generated by both Chrome and Samsung Internet. In both apps we need to open the menu and select the appropriate action.
* `install in app screen` actually installs an app-like thing, i.e. it will show up in the list of installed apps both in the "app screen" but also in the Android settings. It will do the same thing as the shortcut above. But it starts out on the hidden app screen and can be added to the "home screen" if desired. This shortcut does not seem to be an option in Chrome. In Samsung Internet it is just a matter of clicking the appropriate icon in the browser address bar.

That's what I found out by testing on my device. I cannot find any official info on this.

So, this seems to be an additional option in Samsung Internet. I noticed, because HA itself does provide the option to "install in app screen". I know that HA has a service worker. So, I think it is plausible, that this is the requirement we don't fulfil.

akloeckner:
I did some more research and think, what I would like to see is [ambient badging](https://hub.samsunginter.net/docs/ambient-badging/) and installation as a [WebAKP](https://developers.google.com/web/fundamentals/integration/webapks), which come with some [requirements](https://web.dev/install-criteria/),

I also found out, that [Chrome and Edge can install pages as apps](https://web.dev/learn/pwa/installation/#desktop-installation) even on Windows and Linux. I verified with both, that HA itself shows an installation symbol in the address bar while we don't. Maybe this will help with finding out, what is different... 

[...]

PR #838, opened 2022-06-20
fix(webmanifest): add start_url

akloeckner:
This is a requirement for installable web APKs. See #822 and criteria in https://github.com/resoai/TileBoard/issues/819#issuecomment-1059559336 .

### DIFF
--- a/manifest.webmanifest
+++ b/manifest.webmanifest
@@ -1,6 +1,6 @@
 {
   "name": "TileBoard",
-  "start_url": "index.html",
+  "start_url": "./index.html",
   "icons": [
     {
       "src": "favicon.svg",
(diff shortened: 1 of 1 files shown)


manifest.webmanifest
```diff
@@ -1,5 +1,6 @@
 {
   "name": "TileBoard",
+  "start_url": "index.html",
```
rchl, line 3: I'm just now thinking that that maybe this wouldn't be correct in case of using the addon with tileboard exposed on its own port (not ingress). In that case I'm not sure if `index.html` would work. But to be fair, I'm not sure I care... I think that part might be broken anyway due to some HA changes.

I would probably use the `./index.html` for to indicate that it's a relative path

akloeckner: I have that setup (tileboard on its own port and even through a nginx proxy). I tend to open `index.html?jklhdsfkjhsdf` quite often (both on the custom port and through nginx), in order to make sure, the page reloads. So, that should work. And in fact, it did when I installed from my development branch.

